PR #1, opened 2015-01-01
should be   var colCache = require("../colcache");

Timespace7:


m-teslya:
Totally agree. *NIX file system is case-sensitive

guyonroche:
Thanks :-) working on Windows, didn't spot that one

Issue #54, opened 2016-01-05
Corrupted json if xlsx contains in-cell formatting


pvadam:
If a cell contains multiple formats (e.g. different colors in the same cell, which Open XML standard accepts), exceljs splits the text into many cells, therefore the result is corrupted.

Let's say cell A1 contains 3 different words, each in a different format. The parser exports the three words in three json fields, interpreting as ```A1 = [first word]```, ```B1 = [second word]```, ```C3 = [third word]```. Therefore the whole result json is shifted. [See an Open XML spec from MS here.](https://msdn.microsoft.com/en-us/library/office/gg278314.aspx)

PR #55, opened 2016-01-05
Fix issue on in-cell parse (issue #54)

pvadam:
A test file was added to the specs in a base64 format. The PR fixes issue #54 . The sheet looks like this:
image


pvadam:
When do you think it will be merged into the master?

guyonroche:
Looking at this now

guyonroche:
Hmm - there are conflicts. I'd like to investigate this more deeply in any case because if we're going to read rich text, we should also be able to write it too.
I'll add the spec though.

PR #60, opened 2016-01-07
	handle hidden/filtered row attribute

DaSpawn:


DaSpawn:
Mostly completed the streaming worksheet reader, need to work on hyperlinks if time permits. This probably should have been put in new pull request, but not sure how

------------------------------

Working exceljs streaming reader able to handle very large sheets. Tested with workbook 150 cols by 20K rows (product data).


Currently able to read shared strings/workbook/worksheets and return rows as they are read.  Have not tested/written hyperlink handler.  Uses zip streaming handler node-stream-zip which requires file name as input (so can not pipe file handles).  Removed flow control as restricting flow only creates problems.  Removed need for event options as events only emmiteed if listened for 

Listenable Events:
- workbook: error, entry, worksheet, hyperlinks, finished
- worksheet error, row, hyperlink, finished

you can add listeners to the worksheet if needed/wanted data from that sheet

If an on event is not listened for, it is skipped in the stream, so if you are only looking for all the names of the sheets
you would only listen for workbook entry

On workbook events you can add on event handler if needed/wanted to emit row data

```javascript
exceljs = require("exceljs");

var workbook = new exceljs.stream.xlsx.WorkbookReader();

workbook.on('error', function (error) {
        console.log("error", error)
});

workbook.on('entry', function (entry) {
        console.log("entry", entry)
});

workbook.on('worksheet', function (worksheet) {
        console.log("worksheet", worksheet.name)

        worksheet.on('row', function (row) {
                console.log(worksheet.name," row", worksheet)
        });

        worksheet.on('finished', function (rowCount) {
                console.log("worksheet finished, rows: " + rowCount)
        });
});

workbook.on('finished', function () {
        console.log("finished")
});

workbook.read(tmpName);

```

DaSpawn:
did not close this pull, included other changes by accident, hopefully can do new pull

PR #61, opened 2016-01-11
Mostly completed/rewrote the streaming worksheet reader

DaSpawn:
Mostly completed the streaming worksheet reader, need to work on hyperlinks if time permits. 

------------------------------

Working exceljs streaming reader able to handle very large sheets. Tested with workbook 150 cols by 20K rows (product data), but looks as though could handle much more (needs more testing).  Need to add function to pause reading sheet/archive if possible (unsure if really really large sheet could push too many events).  Also need to review other code as blocking functions on row processing are killing the performance (I have to fork a worker process to run this as it hangs my entire application when processing each row).  The blocking functions are killing performance both in this reader and the stream writer. 

Currently able to read shared strings/workbook/worksheets and return rows as they are read.  Have not tested/written hyperlink handler.  Uses zip streaming handler node-stream-zip which requires file name as input (so can not pipe file handles).  Removed flow control as restricting flow only creates problems.  Removed need for event options as events only emmiteed if listened for 

Listenable Events:
- workbook: error, entry, worksheet, hyperlinks, finished
- worksheet error, row, hyperlink, finished

you can add listeners to the worksheet if needed/wanted data from that sheet

If an on event is not listened for, it is skipped in the stream, so if you are only looking for all the names of the sheets
you would only listen for workbook entry

On workbook events you can add on event handler if needed/wanted to emit row data

```javascript
exceljs = require("exceljs");

var workbook = new exceljs.stream.xlsx.WorkbookReader();

workbook.on('error', function (error) {
        console.log("error", error)
});

workbook.on('entry', function (entry) {
        console.log("entry", entry)
});

workbook.on('worksheet', function (worksheet) {
        console.log("worksheet", worksheet.name)

        worksheet.on('row', function (row) {
                console.log(worksheet.name," row", worksheet)
        });

        worksheet.on('finished', function (rowCount) {
                console.log("worksheet finished, rows: " + rowCount)
        });
});

workbook.on('finished', function () {
        console.log("finished")
});

workbook.read(tmpName);

```

### DIFF
--- a/lib/stream/xlsx/workbook-reader.js
+++ b/lib/stream/xlsx/workbook-reader.js
@@ -20,210 +20,245 @@
  * THE SOFTWARE.
  * 
  */
-"use strict";
+ "use strict";
 
-var fs = require("fs");
-var events = require("events");
-var Stream = require("stream");
-var Promise = require("bluebird");
-var _ = require("underscore");
-var unzip = require("unzip");
-var Sax = require("sax");
+ var fs = require("fs");
+ var async = require("async");
+ var events = require("events");
+ var Stream = require("stream");
+ var Promise = require("bluebird");
+ var _ = require("underscore");
+ var StreamZip = require("node-stream-zip");
+ var Sax = require("sax");
 
-var utils = require("../../utils/utils");
-var Enums = require("../../enums");
-var StreamBuf = require("../../utils/stream-buf");
-var StutteredPipe = require("../../utils/stuttered-pipe");
-var FlowControl = require("../../utils/flow-control");
-var AutoDrain = require("../../utils/auto-drain");
+ var utils = require("../../utils/utils");
+ var Enums = require("../../enums");
+ var StreamBuf = require("../../utils/stream-buf");
+ var StutteredPipe = require("../../utils/stuttered-pipe");
+ var FlowControl = require("../../utils/flow-control");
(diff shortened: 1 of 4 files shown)


DaSpawn:
DO NOT MERGE YET

There appears to be an issue with reading the sheet and column alignment/shared string alignment that is still unresolved and making reading potentially incorrect (some sheets I tried work fine, others shift randomly by one column, and the field object data directly from parsed rows is the wrong column.  I thought it was my own code but I have seen this inconsistency/bug before and never pinpointed)

alangunning:
Keep up the great work :+1:

DaSpawn:
I have completely rewritten the Excel streaming reader from scratch because I was unable to find the column shifting bug (in the past I worked with a PHP excel reader that had a similar column shifting issue and was due to excel being 1 based indexed instead of zero that really complicates things)

kmorte:
Hi Brian,

what is the state of the read stream performance?
You talked abou 20k rows, but what about 1M lines?
I'm currently trying to manage this load and with ExcelJs it just takes too many hours. Maybe I'm doing something wrong.

Thanks!

DaSpawn:
I created this module: https://github.com/daspawn/xlsx-stream-reader

Should take care of that many rows without issue

PR #102, opened 2016-04-11
bugfix: numFmtHash's wrong hash-key makes too many <numFmt> tags for …

Youngbae-Jeon:
…the same format in styles.xml

### DIFF
--- a/lib/xlsx/stylemanager.js
+++ b/lib/xlsx/stylemanager.js
@@ -360,7 +360,7 @@ utils.inherits(StyleManager, events.EventEmitter, {
     _addNumFmt: function(numFmt) {
         // called during parse
         this.numFmts.push(numFmt);
-        this.numFmtHash[numFmt.id] = numFmt;
+        this.numFmtHash[numFmt.formatCode] = numFmt;
         return numFmt.id;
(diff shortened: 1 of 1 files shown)


Youngbae-Jeon:
Too many number formats seems to makes the excel file unreadable.
https://support.microsoft.com/en-us/kb/213904

keeneym:
I was having the same problem. This fix worked great.

FYI if it helps here was the specific error I was getting when validating sheets with too much style:

`Location of error: { 0}/x:worksheet[1]/x:sheetData[1]/x:row[4431]/x:c[11]
Error description: The attribute 's' has invalid value '66473'. The MaxInclusive constraint failed. The value must be less than or equal to 65490.`

So it looks like the max for xlsx is about 65490.

surjikal:
This is fantastic, thanks!

guyonroche:
I think this fix is already part of master, @Youngbae-Jeon - can you confirm?

Youngbae-Jeon:
I have tested with the last commit on Aug 16, 2016 and found there are no more problems like this.
So I'm closing this pull request. Thank you.

PR #103, opened 2016-04-11
some logs not commented out

timruffles:
These log messages appear in our automatic tests.

I notice you've used commented logs throughout. Would it be useful to you if I switched it over to `debug` so you can turn them on/off with env vars rather than comments? It's easy to forget to re-comment them, which I assume is why they're here!


guyonroche:
Workbook Reader is not really finished yet but I'll merge this...

PR #105, opened 2016-04-12
Print settings

jackkum:
Ability to control print settings.


Loksly:
Thanks a lot for your contribution.

PR #137, opened 2016-06-17
Fall back to JSON.stringify() if unknown Cell.Type

wulfsolter:


guyonroche:
Not sure I agree with this - what good would it do to add this cell-type?
How would it be rendered in the file output?

wulfsolter:
My use case for adding cell type is I'm attempting to generate XLS' of varied external data. Often enough a column of objects or other data pops up that isn't one of the ExcelJS types, and this gives a fall back output of JSON.stringified() which I think is better than breaking.

PR #138, opened 2016-06-17
Not explode if model is undefined

wulfsolter:


guyonroche:
I believe this has already been addressed in v0.2.9
Can you confirm?

wulfsolter:
It has! And you've done some pretty amazing work over the last few weeks! Thanks heaps :)

PR #152, opened 2016-07-22
fix broken tables at README

haoliangyu:


guyonroche:
Those tables were fixed before I managed to see this PR.
Thanks for the contribution though

PR #196, opened 2016-10-13
fix typo on sheet view showGridLines option

gadiaz1:
According to the [documentation](https://msdn.microsoft.com/en-us/library/documentformat.openxml.spreadsheet.sheetview.showgridlines(v=office.14).aspx) the showGridLines var in worksheet views options is misspelled. It should be "showGridLines", not "showGridlines".

I changed all the files, and now it is working.


dimfeld:
@gadiaz1 Just today I needed to use showGridLines and ran into the same problem you fixed here. Thanks for putting this together!

Issue #163, opened 2016-08-03
Error while using xslx readFile method


carmelo12341:
```
var spans = node.attributes.spans.split(':').map(function(span) { return parseInt(span); });
                                       ^

TypeError: Cannot read property 'split' of undefined
    at utils.inherits.parseOpen (/home/carmelo12341/dev/nodejs/excel/node_modules/exceljs/lib/xlsx/xform/sheet/row-xform.js:97:40)
    at utils.inherits.parseOpen (/home/carmelo12341/dev/nodejs/excel/node_modules/exceljs/lib/xlsx/xform/list-xform.js:71:31)
    at utils.inherits.parseOpen (/home/carmelo12341/dev/nodejs/excel/node_modules/exceljs/lib/xlsx/xform/sheet/worksheet-xform.js:109:19)
    at SAXStream.<anonymous> (/home/carmelo12341/dev/nodejs/excel/node_modules/exceljs/lib/xlsx/xform/base-xform.js:65:14)
    at emitOne (events.js:77:13)
    at SAXStream.emit (events.js:169:7)
    at Object.me._parser.(anonymous function) [as onopentag] (/home/carmelo12341/dev/nodejs/excel/node_modules/exceljs/node_modules/sax/lib/sax.js:245:15)
    at emit (/home/carmelo12341/dev/nodejs/excel/node_modules/exceljs/node_modules/sax/lib/sax.js:615:33)
    at emitNode (/home/carmelo12341/dev/nodejs/excel/node_modules/exceljs/node_modules/sax/lib/sax.js:620:3)
    at openTag (/home/carmelo12341/dev/nodejs/excel/node_modules/exceljs/node_modules/sax/lib/sax.js:801:3)
    at Object.write (/home/carmelo12341/dev/nodejs/excel/node_modules/exceljs/node_modules/sax/lib/sax.js:1181:11)
    at SAXStream.write (/home/carmelo12341/dev/nodejs/excel/node_modules/exceljs/node_modules/sax/lib/sax.js:227:16)
    at Entry.ondata (_stream_readable.js:542:20)
    at emitOne (events.js:77:13)
    at Entry.emit (events.js:169:7)
    at readableAddChunk (_stream_readable.js:153:18)
```
In node 6.3.1 I get an additional error:
 ```
(node:2898) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
```
The error disappears if I use the LTS version (4.4.7)
This is my code:
[...]

PR #203, opened 2016-10-20
Fix issue #163: the "spans" attribute of the row element is optional

arturas-vitkauskas:
The "spans" attribute is not required in the "row" element. Source of information: https://msdn.microsoft.com/en-us/library/office/documentformat.openxml.spreadsheet.row.aspx

If we create a Google Sheets spreadsheet and download it as XLSX, then the "spans" attributes do not appear in the "row" elements of the worksheet.

Attached is a test xlsx file, which causes the "Cannot read property 'split' of undefined" error when trying to read it with the exceljs package.
[test-issue-163.xlsx](https://github.com/guyonroche/exceljs/files/542029/test-issue-163.xlsx)


tkirda:
Looks good, waiting for the merge and NPM publish.

PR #205, opened 2016-10-25
utc Offsset to excelToDate and dateToExcel

skumarnk2:
updated excelToDate and dateToExcel to consider utc Offset.
If my machine's timezone is GMT+0530 and my date in Date object is 20-Oct-2016 00:00:00 then in excel file written with exceljs it is shown as 19-Oct-2016 06:30:00 PM.

### DIFF
--- a/lib/utils/utils.js
+++ b/lib/utils/utils.js
@@ -119,10 +119,12 @@ var utils = module.exports = {
   },
   inherits: inherits,
   dateToExcel: function(d) {
-    return 25569 + d.getTime() / (24 * 3600 * 1000);
+    return 25569 + (d.getTime() - d.getTimezoneOffset()*60*1000) / (24 * 3600 * 1000);
   },
   excelToDate: function(v) {
-    return new Date((v - 25569) * 24 * 3600 * 1000);
+    var d = new Date((v - 25569) * 24 * 3600 * 1000);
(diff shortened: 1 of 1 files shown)


guyonroche:
@skumarnk2 - I'm not comfortable merging this change as it will quite likely break other users' code. I also don't think this is the correct solution to this as there is no information in the getTimezoneOffset() function to infer what time-zone the Date object was built from.

The getTime() function returns an epoch time (which is UTC) and from what I've seen, Excel's datetimes are zoneless so to try and compensate like this seems wrong.

As a work-around, you should create your date objects like this...

```javascript
var date  = new Date(Date.utc(year,  month, day,  hour, minute, second, ms));
```

When you add dates like this into Excel, you will see the hours and minutes that you want.

skumarnk2:
@guyonroche, thanks for your detailed analysis.

Since Excel's datetimes are zoneless, we should ignore the zone part when writing date object to Excel and should use system's default zone when reading from excel.

getTimezoneOffset() gives the timezone offset associated with date object.

the way i suggested, is working similar to 
<code>var date = new Date(year,  month, day,  hour, minute, second, ms);</code>
for reading and just reverse for writing. It is not influencing system to not pick default time zone.

if you are still not satisfied, can you tell me in which scenario you are thinking it could go wrong.

PR #214, opened 2016-11-01
Fixed tests

oferns:
Test all passing


guyonroche:
Hi @oferns 
Not sure on the purpose of this PR - the only significant changes are in the Row - builds a model unit test, changing the row from sparse (i.e. missing column 'C') to contiguous.
Also - the tests pass already.
Could you provide explaination?
Thanks,
Guyon

PR #217, opened 2016-11-04
Stream reader fixes

kturney:
These are the fixes I needed to make to get streaming xlsx reading working.


kturney:
Hi @guyonroche.
Thanks for the quick merge.
Do you think this could make it in a release soon?
Thanks again.

kturney:
whoops! nevermind. Missed that you already did :-)

PR #225, opened 2016-11-11
Fix output order of Sheet Properties

keeneym:
I noticed that when using both "tabColors" and "sheetViews" the resulting xlsx files end up being corrupt when trying to open in Microsoft Excel. After playing around a bit in the zip files I realized Excel is pretty picky about element order in the xml files, and it prefers to have the Sheet Properties before the Sheet Views. 

Here is some example code that is corrupt before my change and opens fine after:

```var Excel = require('../exceljs/excel');
var wb = new Excel.stream.xlsx.WorkbookWriter({filename: "tabColorTest.xlsx", useSharedStrings: true});
var ws = wb.addWorksheet('blort', {
    properties: {
        tabColor: {argb: 'FFFF0000'}
    },
    views: [
        {
            state: 'frozen',
            ySplit: 1,
            topLeftCell:"A2"
        }
    ]
});

ws.getCell('A1').value = "frozenHeader";
ws.getCell('A2').value = "tabColorTest";
wb.commit();
```


keeneym:
@guyonroche Was I supposed to create an Issue for this? I'm a bit new to github, I'm not sure if you see incoming Pull Requests.

guyonroche:
@keeneym no need for an issue. Thanks for the contribution.

Issue #219, opened 2016-11-06
1904 Dates


Agoraios:
In Excel if you put in a date and set the date system to 1904, exceljs gives the wrong date by 4 years

guyonroche:
Should be fixed by [Start on support for 1904 based dates #237](https://github.com/guyonroche/exceljs/pull/237) and released in 0.2.38

Agoraios:
Thank you very much @holm !

PR #237, opened 2016-12-05
Start on support for 1904 based dates

holm:
This is just the bare minimum start for date 1904 support, but it is as much as I have time for right now.

It should apply to the parsing and formatting of dates, but I have not been able to figure out how to get the property passed into the date handling, so it can be accounted for.

It will also require more tests before it is ready to be merged in.

References #219


guyonroche:
@holm - thanks for this. I finished the changes and it looks good now.
I changed the date1904 property value from 1 to true - seems more logical ;-)

holm:
Makes sense, I also considered it, but just stuck to what they used in the file format. Thanks for finishing this up and getting it in.

Issue #252, opened 2017-01-12
Data validation formulae undefined bug


jayflo:
Receiving the following error:

```
TypeError: Cannot read property 'forEach' of undefined
      at node_modules/exceljs/lib/xlsx/xform/sheet/data-validations-xform.js:114:23
      at node_modules/exceljs/node_modules/lodash/lodash.js:4944:15
      at baseForOwn (node_modules/exceljs/node_modules/lodash/lodash.js:3001:24)
      at node_modules/exceljs/node_modules/lodash/lodash.js:4913:18
      at Function.forEach (node_modules/exceljs/node_modules/lodash/lodash.js:9359:14)
      at render (node_modules/exceljs/lib/xlsx/xform/sheet/data-validations-xform.js:83:9)
      at render (node_modules/exceljs/lib/xlsx/xform/sheet/worksheet-xform.js:119:30)
      at node_modules/exceljs/lib/xlsx/xlsx.js:342:24
      at Array.forEach (native)
      at node_modules/exceljs/lib/xlsx/xlsx.js:340:24
      at Promise._execute (node_modules/bluebird/js/release/debuggability.js:300:9)
      at Promise._resolveFromExecutor (node_modules/bluebird/js/release/promise.js:481:18)
      at new Promise (node_modules/bluebird/js/release/promise.js:77:14)
      at Object.addWorksheets (node_modules/exceljs/lib/xlsx/xlsx.js:334:12)
      at node_modules/exceljs/lib/xlsx/xlsx.js:418:21
      at tryCatcher (node_modules/bluebird/js/release/util.js:16:23)
      at Promise._settlePromiseFromHandler (node_modules/bluebird/js/release/promise.js:510:31)
      at Promise._settlePromise (node_modules/bluebird/js/release/promise.js:567:18)
      at Promise._settlePromise0 (node_modules/bluebird/js/release/promise.js:612:10)
      at Promise._settlePromises (node_modules/bluebird/js/release/promise.js:691:18)
      at Promise._fulfill (node_modules/bluebird/js/release/promise.js:636:18)
      at PromiseArray._resolve (node_modules/bluebird/js/release/promise_array.js:125:19)
      at PromiseArray._promiseFulfilled (node_modules/bluebird/js/release/promise_array.js:143:14)
      at Promise._settlePromise (node_modules/bluebird/js/release/promise.js:572:26)
      at Promise._settlePromise0 (node_modules/bluebird/js/release/promise.js:612:10)
[...]

PR #253, opened 2017-01-12
[BUGFIX] data validation formulae undefined

jayflo:
Source: #252 

* make sure formulae is an array
* add example to spec


guyonroche:
Nice catch! I hadn't considered validations that didn't have formulae since Excel doesn't let you commit the modal if you don't fill them in (except for 'any', which I'd also missed).
However Excel will load a decimal validation without formulae so it's ok to render such.

PR #255, opened 2017-01-19
added a (maybe partial) solution to issue 99. i wasn't able to create an appropriate test

misleadingTitle:
With my partially implemented code I can reproduce the issue: the generated file shows as corrupted in Excel (version 2016). Analyzing the corrupted sheet XML I saw that a tag was closed but never opened, so I tried to fix that so that Excel shows an empty sheet (as should be) and doesn't complain.
Unfortunately I was no able to create an appropriate test so I created a failing stub for one.


guyonroche:
Thanks for the contribution
I added a unit-test to catch the issue and dropped your integration test stub

PR #256, opened 2017-01-25
Sheets with hyperlinks and data validations are corrupted

simon-stoic:
I was trying to convert a file that had both an hyperlink and data validations. For some obscure reason, Excel didn't want to open the resulting file. Putting the links after the data validations in the sheet's XML fixed the issue.
I'm not familiar enough with exceljs to write a test for this issue.


guyonroche:
Good catch. Don't worry, I'll add a unit test for this

PR #259, opened 2017-01-31
Add support for identifier

holm:
I ran into a similar issue to https://github.com/guyonroche/exceljs/issues/176, just for dc:identifier. I found the property in the spec, so I added it in this PR. Unfortunately I haven't been able to receive the file from the customer that ran into this (I only have the stacktrace), but I am trying to get it so I can create a test case.


guyonroche:
Cool - thanks.
If you do manage to get the file and write a test-case, please do add another PR for it :-)

PR #267, opened 2017-02-20
Minor fixes for stream handling

holm:
I upgraded from 0.2.38 to 0.2.46 today, and a few of our testcases failed. I tracked it down to two issues:

* The catch handler on the zip parsing was missing, and also wasn't propagating the error via the event emitter.
* The workbook writer was resolving the write when the zip-stream was done, not when the underlying stream was done. This meant that the write could resolve, while f.ex. a file was still being written to disk.

### DIFF
--- a/lib/stream/xlsx/workbook-writer.js
+++ b/lib/stream/xlsx/workbook-writer.js
@@ -319,7 +319,7 @@ WorkbookWriter.prototype = {
       self.stream.on('error', function(error){
         reject(error);
       });
-      self.zip.on('finish', function(){
+      self.stream.on('finish', function(){
         resolve(self);
       });
       self.zip.on('error', function(error){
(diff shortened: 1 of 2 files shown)


neuralp:
Thank you for this simple fix.  I was just trying to figure out why my catch wasn't working while passing bad files to the readFile function and behold, you have the answer!

guyonroche:
@holm - you wouldn't be able to add one of those failing test cases (or a smaller version of one) would you?
This (or something related to it) has been a recurring issue and it would be good to prevent it from happening again.

holm:
I can try to find the time next week. I just needed it fixed quite urgently for one of our customers, but I should have time soon.

guyonroche:
@holm - Just looking into this now, I hope to publish this weekend. I just noticed the change in zip-stream removed the .catch() on entry.async(), was that necessary or is it just that entry.async() doesn't reject?

holm:
Yeah, that looks like a mistake. Should most likely emit the error there also.

guyonroche:
@holm  Cool - thanks.
I've restored the other catch.
All tests passing with this and the updated lib versions so I've just published => 0.3.1.

holm:
Thanks a lot for all your work on this library, it's pretty great!

PR #294, opened 2017-03-27
Support error references in cell ranges

holm:
This PR adds support for decoding references with an error. It does this by using a regex to find the sheetname, if any, and handle the optional quotes. I wish JS regex had named groups, in which case this would be a bit nicer.

### DIFF
--- a/lib/utils/col-cache.js
+++ b/lib/utils/col-cache.js
@@ -165,10 +165,10 @@ var colCache = module.exports = {
 
   // convert [sheetName!][$]col[$]row[[$]col[$]row] into address or range structures
   decodeEx: function(value) {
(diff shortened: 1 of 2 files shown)


Rycochet:
Might be worth using non-capturing groups to clean up the regex a little and make it obvious which parts are important - /`(?:...)/`

holm:
I added the non-capturing groups, and added a little bit of comments.

holm:
Anything I can do to improve this PR to get it merged?

guyonroche:
@holm - sorry to take so long - my head was buried in the image branch. Just published this today

holm:
Thanks a lot!

PR #297, opened 2017-04-03
Issue with copied cells

muscapades:
This demonstrates a problem where cells that are copied as a block does not appear to have formulas, only values.

I suspect that Excel has some way of saving "similar cells" that exceljs does not understand?

The attached fibonacci.xslx has the problem. Rows 1-3 where created by hand.
Rows 4 to 19 where created by copying A3:B3 in one copy-paste operation.

The first row in the new block, A4:B4, is fine.

All the rest of the rows, A5:B19, appear in exceljs to have only values, not formulas.

I can reproduce this problem by copying blocks of cells in other spreadsheets too. The first row is okay, the rest shows the problems. 
Copying the rows one by one is a workaround, but not reliable. (A sheet created like this worked, but then when I added a range name to a cell, many cells now start showing the problem again.)


muscapades:
More debugging info:
If I add this logging in cell-xform.parseOpen, I see that the cells are indeed referencing each other.
A4 has `{ t: 'shared', ref: 'A4:A19', si: '0' }` in the <f> tag, and
A5 has `{ t: 'shared', si: '0' }`

I am reading this "A4 is the prototype for all of A4:A19, call that si:0" and "A5 shares the prototype si:0"

muscapades:
Right. This is in the ECMA specification, like pointed out here: 
http://stackoverflow.com/a/18375178, and I found it in the spec by following the link too.
So the formulas should have the same representation in R1C!-reference notation...

muscapades:
The last commit is my attempt at a fix; your input is welcome

guyonroche:
@muscapades - Thanks for finding this. I'll review and publish it as soon as I can.

muscapades:
@guyonroche I have a working solution for calculating the "copied" formula too - I'll wrap that up today, I think. I was thinking I should do that in another branch so that you can take your pick.

muscapades:
Consider https://github.com/guyonroche/exceljs/pull/300 instead - it also adds the formula as it would be for the cell.

(It can be considered whether it should then just be a normal FormulaValue again, and forget about SharedFormulaValue again)

Issue #197, opened 2016-10-13
Is it possible to define automatic column filters


i0natan:
Hi,

Thanks for the great work.

Question: is it possible to define per-sheet an automatic filter for all/some columns? in other words, when the user will open the sheet, automatically Excel will display filters on the column headers

thanks,
Yoni Goldberg

Issue #112, opened 2016-04-21
Filters at top of page


mpushkix:
Hi,
Plaese, tell me, is it possible to add a filter for the headers with your library?

bluedge:
+1 would be a great feature to add

PR #306, opened 2017-04-21
Added support for auto filters

C4rmond4i:
This pull request addresses the feature requests #197 and #112.


AhmadMurtaza1:
What I want is to retrieve only filtered rows from csv but It will shows or results all rows in console.log

```
    var file = './uploads/attachFile-1508833322746.csv';
    workbook.csv.readFile(file).then(function(worksheet) {
        worksheet.autoFilter = {
            from: {
                row: 3,
                column: 1
            },
            to: {
                row: 5,
                column: 12
            }
        }
        worksheet.eachRow({ includeEmpty: true }, function(row, rowNumber) {
            console.log(JSON.stringify(row.values));
        });
    })
```
Looking forward to hear from you.
Thanks in Advance 👍

PR #320, opened 2017-05-18
Add printTitlesRow Support

psellers89:


psellers89:
Done. Let me know if it is ok now.

PR #343, opened 2017-06-21
Issue 296

holly-weisser:
Let new lines through xmlEncode function. See github issue 296 (https://github.com/guyonroche/exceljs/issues/296).

### DIFF
--- a/lib/utils/utils.js
+++ b/lib/utils/utils.js
@@ -79,7 +79,7 @@ var utils = module.exports = {
   },
   xmlEncode: function(text) {
     // eslint-disable-next-line no-control-regex
-    return text.replace(/[<>&'"\x7F\x00-\x08\x0A-\x0C\x0E-\x1F]/g, function(c) {
+    return text.replace(/[<>&'"\x7F\x00-\x08\x0B-\x0C\x0E-\x1F]/g, function(c) {
       switch (c) {
         case '<': return '&lt;';
         case '>': return '&gt;';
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "exceljs",
-  "version": "0.4.12",
+  "version": "0.4.13",
```
guyonroche: Bumping the package version is managed by the release scripts (npm version patch, etc). Please can you revert this line

lib/utils/utils.js
```diff
@@ -79,7 +79,7 @@ var utils = module.exports = {
   },
   xmlEncode: function(text) {
     // eslint-disable-next-line no-control-regex
-    return text.replace(/[<>&'"\x7F\x00-\x08\x0A-\x0C\x0E-\x1F]/g, function(c) {
+    return text.replace(/[<>&'"\x7F\x00-\x08\x0B-\x0C\x0E-\x1F]/g, function(c) {
```
guyonroche: Could you also add a unit test for this (in spec/unit/utils/utils.spec.js to ensure \x0a is retained

holly-weisser: Weird that the previous PR that was supposed to fix this same issue wasn't required to have a unit test.

guyonroche: I had a lot more spare time then. Thanks for doing this - I'll publish tonight.

PR #364, opened 2017-07-19
Fix #345 TypeError: Cannot read property 'date1904' of undefined

### LINKED ISSUE #345: TypeError: Cannot read property 'date1904' of undefined
export xlsx by macOS Numbers app, then i use：

```
workbook.xlsx.readFile('template.xlsx');
```

```
TypeError: Cannot read property 'date1904' of undefined
      at module.exports.reconcile (/Users/bukas/web/api/node_modules/exceljs/dist/es5/xlsx/xlsx.js:108:33)
      at /Users/bukas/web/api/node_modules/exceljs/dist/es5/xlsx/xlsx.js:316:14
      at <anonymous>
```

Diluka:
cant open xlsx saved from Numbers
model.properties is undefined

### DIFF
--- a/lib/xlsx/xform/book/workbook-xform.js
+++ b/lib/xlsx/xform/book/workbook-xform.js
@@ -127,7 +127,7 @@ utils.inherits(WorkbookXform, BaseXform, {
       case 'workbook':
         this.model = {
           sheets: this.map.sheets.model,
-          properties: this.map.workbookPr.model,
+          properties: this.map.workbookPr.model || {},
           views: this.map.bookViews.model
         };
         if (this.map.definedNames.model) {


lib/xlsx/xform/book/workbook-xform.js
```diff
@@ -127,7 +127,7 @@ utils.inherits(WorkbookXform, BaseXform, {
       case 'workbook':
         this.model = {
           sheets: this.map.sheets.model,
-          properties: this.map.workbookPr.model,
+          properties: this.map.workbookPr.model || {},
```
guyonroche: Looks good - could you also add a unit test that catches this issue?
in: spec/unit/xlsx/xform/book/workbook-xform.spec.js
best approach woud be making copies of the spec/unit/xlsx/xform/book/data/book.1.* files and adding a new expectation

Diluka: Sorry. No, actually I don't know how to add the unit test. 
I just fixed an `.xlsx` file parsed with its `this.map.workbookPr.model` being `undefined` causing an null reference error. 
This because the file itself having a missing property and `exceljs` supposing it have.

guyonroche: OK, I have a little time now, I'll do it

PR #403, opened 2017-09-15
Guard null model fields - fix and tests

thecjharries:
I ran into an issue with the [`reconcile` method](https://github.com/guyonroche/exceljs/blob/master/lib/xlsx/xform/drawing/two-cell-anchor-xform.js#L108) of `TwoCellAnchorXform`. I can't provide the spreadsheet to give you a working demo of the issue, but I did find the problem and built a unit test to display the issue. Basically, `model.picture` might be null, which throws an error. Once I added the guards, I was able to load the workbook as usual.

### DIFF
--- a/lib/xlsx/xform/drawing/two-cell-anchor-xform.js
+++ b/lib/xlsx/xform/drawing/two-cell-anchor-xform.js
@@ -106,14 +106,16 @@ utils.inherits(TwoCellAnchorXform, BaseXform, {
   },
 
   reconcile: function(model, options) {
-    var rel = options.rels[model.picture.rId];
-    var match = rel.Target.match(/.*\/media\/(.+[.][a-z]{3,4})/);
-    if (match) {
-      var name = match[1];
-      var mediaId = options.mediaIndex[name];
-      model.medium = options.media[mediaId];
+    if (model.picture && model.picture.rId) {
+      var rel = options.rels[model.picture.rId];
+      var match = rel.Target.match(/.*\/media\/(.+[.][a-z]{3,4})/);
+      if (match) {
+        var name = match[1];
+        var mediaId = options.mediaIndex[name];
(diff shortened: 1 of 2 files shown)


lib/xlsx/xform/drawing/two-cell-anchor-xform.js
```diff
@@ -106,14 +106,16 @@ utils.inherits(TwoCellAnchorXform, BaseXform, {
...
+      var rel = options.rels[model.picture.rId];
+      var match = rel.Target.match(/.*\/media\/(.+[.][a-z]{3,4})/);
+      if (match) {
+        var name = match[1];
+        var mediaId = options.mediaIndex[name];
+        model.medium = options.media[mediaId];
+      }
     }
-    if (Number.isInteger(model.tl.row) && Number.isInteger(model.tl.col) && Number.isInteger(model.br.row) && Number.isInteger(model.br.col)) {
+    if (model.tl && Number.isInteger(model.tl.row) && Number.isInteger(model.tl.col) && Number.isInteger(model.br.row) && Number.isInteger(model.br.col)) {
```
Rycochet: Is it worth adding a test for `model.br` too? Obviously I don't know if it's possible for it to be null or not, but the code doesn't give any indication either way.

thecjharries: Great point. I went ahead and added a `br` guard and test. I should have done that originally.

thecjharries: I started with the `tl` guard because I was testing with `model = {}`, which threw an error on that conditional. I didn't have the time to debug everything enough to figure out where `model` was coming from and why its `picture` was `null`, so I'm as in the dark about whether or not `br` can be `null` without a `null` `tl` as you are. It doesn't hurt, so there's no reason not to.

PR #466, opened 2017-12-19
Don't break when loading an Excel file containing a chartsheet

papandreou:
I ran into a problem where exceljs crashed on a customer-uploaded excel file containing chartsheets. It's an entirely new concept to me, and it looks like exceljs doesn't understand those at the moment, so I chose to do a minimal fix so we don't crash, at least :)

Test case found here: http://www.excel-easy.com/examples/excel-files/chart-sheet.xlsx


AlameerAshraf:
It's still breaking when the excel file contains a chart sheet :( any ideas !

yogithesymbian:
> It's still breaking when the excel file contains a chart sheet :( any ideas !

where is documentation to show a chart ?

PR #478, opened 2018-01-18
Round Excel date to nearest millisecond when converting to javascript date

bjet007:
Hi, 

while I was trying to test the generation of our excel file using your library, I found out that some date were not matching what I was expecting.

Opening my file in excel seem to display the correct value, but reloading it with the library and then comparing the excel value to what I was expecting was not always matching. By example writing the following date 2017-11-15T17:00:00.000Z to an excel file and then re-reading it would return 2017-11-15T16:59:59.999Z.  

Since excel already round value to the millisecond when showing the value and a javascript date doesn't have precision under a millisecond, I would like to propose that we round the value to the nearest millisecond when reading an excel file.

I haven't found any mention of precision in the spec file, but I think it would be better than truncating the value as it would at least keep the same value when writing and reading a file.

### DIFF
--- a/lib/utils/utils.js
+++ b/lib/utils/utils.js
@@ -64,7 +64,8 @@ var utils = module.exports = {
     return 25569 + (d.getTime() / (24 * 3600 * 1000)) - (date1904 ? 1462 : 0);
   },
   excelToDate: function(v, date1904) {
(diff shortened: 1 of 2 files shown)


bjet007:
Had a bug in my test while running 'npm run preversion'

PR #485, opened 2018-01-29
Bug fix: wb.worksheets/wb.eachSheet caused getWorksheet(0) to return sheet

mah110020:
wb._worksheets is an array with an empty first element. Calling wb.worksheets or wb.eachSheet called _worksheets.sort() which mutated the array and shifted the sheets into the empty element slot. I am not sure that the mutation was intentional or necessary. My solution no longer mutates _worksheets.

Also added test.


guyonroche:
Nice find - just publishing it now... Thanks @mah110020

Issue #238, opened 2016-12-08
Typescript type declarations for exceljs


madoar:
I want to use exceljs in a typescript project of mine. When using exceljs in a typescript project, it would be nice if there is a type declaration file (d.ts file) out there for exceljs to prevent programming errors.

Is there a typescript type declaration file for exceljs?

Rycochet:
I've started putting together a .d.ts for this - https://gist.github.com/Rycochet/ebd8521c56059ec440339098b18a58c7

Please note that it's still a WIP - I've grabbed comments from the README or the source where possible, and tried to make sure types are correct as much as is needed (so there may be some placeholder "any" types around).

Quite happy for this to join the repo - but want to get it a bit more filled out first, putting it out there for people to use (and hopefully feedback on stuff I've missed). My needed target is browser - so it'll almost definitely be missing Node related bits.

ilya-payrange:
Having .d.ts file will be really great!

Meanwhile I used this method (presented in #335)

In project tree:
`npm install exceljs --save`

In the component.ts:
`import * as ExcelJs from 'exceljs/dist/exceljs.min.js';`

Do not try to do dependency injection into constructor() (it requires proper .d.ts). Instead just grab the ExcelJs:

`var workbook = new ExcelJs.Workbook();`

Works for me. 

P.S. When I was trying `var workbook = new ExcelJs.Workbook();` with `import ExcelJs from 'exceljs;`, app crashed.

Rycochet:
@ilya-payrange My [gist](https://gist.github.com/Rycochet/ebd8521c56059ec440339098b18a58c7) *is* a .d.ts file for it, probably about 90% complete - moved on to other projects at work so don't know if I'll get a chance to update it at all any time soon though...

Issue #177, opened 2016-09-02
Typings


PRR24:
Exceljs feels like very well though out and structured solution and I think it does deserve Typescript definitions.
Please consider.

ZYahia:
I really need typescript definition for this library , please consider it

Rycochet:
Continued in #238

Diluka:
missing `index.d.ts` in latest release? @guyonroche

PR #490, opened 2018-02-03
add type definition

taoqf:


Rycochet:
Should reference to #238 and #177 here, and maybe compare contents of my gist linked in it.

taoqf:
I just want help, and I am using this project in my work.

kachkaev:
Having typings for this library would be awesome! @guyonroche 🙏

package.json
```diff
@@ -98,4 +99,4 @@
     "LICENSE",
     "README.md"
   ]
-}
+}
```
kachkaev: ouch 😆

taoqf: Sorry for the formatter

guyonroche:
Thanks @taoqf for doing this!
I'm not the hugest fan of types in JS but I know a lot of people are :-) so this will be a great addition

Purus:
How should I use this typings for my Typescript application?

taoqf:
Just install this lib will be OK, after new version if published.

taoqf:
ask @guyonroche

ahaverty:
@taoqf @guyonroche It looks like this is in the repo and released 👏 

But on v1.0.1 `import * as ExcelJs from 'exceljs';` isn't giving me any types as expected.
I can't find a index.d.ts in the dist after installing either
Has anyone gotten it to work? 🤔

taoqf:
@guyonroche , please put index.d.ts in release.

grbspltt:
@ahaverty I found the typings in the repo, copied and pasted into the node_modules/exceljs folder.  Not a great solution but works until the release is fixed.

spacem:
Before this change I was using types from https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/exceljs
ie. installed with `npm install @types/exceljs`

Now I updated to the latest exceljs and the 'official' types from this merge don't seem so good. To give one example the font interface says that font requires all the parameters where I believe actually they should be optional?

kachkaev:
@spacem could you please submit a PR that would fix the typing issues you are facing?

taoqf:
@spacem anything else?

PR #495, opened 2018-02-07
Add correct parsing for inline strings

robscotts4rb:
Inline strings that previously met the XML spec, caused  the parsing of the files to fail.

### DIFF
--- a/lib/xlsx/xform/sheet/cell-xform.js
+++ b/lib/xlsx/xform/sheet/cell-xform.js
@@ -251,6 +251,10 @@ utils.inherits(CellXform, BaseXform, {
         this.currentNode = 'v';
         return true;
 
+      case 't':
+        this.currentNode = 't';
+        return true;
+
       default:
         return false;
     }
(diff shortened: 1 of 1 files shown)


robscotts4rb:
@guyonroche I couldn't find the correct place to add tests around parsing of a file. If you can point me in the correct direction I am happy to add them

guyonroche:
@robscotts4rb sorry it's taken me so long to get to this.
If you could add some tests that would be fantastic. The best file would be spec/unit/xlsx/xform/sheet/cell-xform.spec.js

Would this mean we should be writing inline strings as <c t="inlineStr"><t> ... </t></c> ?
Or do you think that might cause more problems than it solves?

linguamatics-pdenes:
@guyonroche I tried to something similar in #501 and added a very basic unit test in there -- hope it makes sense!

guyonroche:
@robscotts4rb thanks for the contribution - I'm going to merge #501 and close this one as they are both working to the same end

PR #496, opened 2018-02-10
Tests against modern node versions

amilajack:


guyonroche:
@amilajack this change has already been applied so I'll close this.
Thanks for the contribution though

PR #501, opened 2018-02-20
Add support for inlineStr, including rich text

### LINKED ISSUE #409: Rich text formatting is not rendered when using the Streaming I/O
Hello there,

In my application I'm building an Excel document row by row and stream each row when it's done. One cell has rich formatting - some words have specific colours. I think I'm doing the right thing but in the Excel document the cell has the value `[Object object]`.
I guess this is the `toString()` of the object instead of the correct rendering.

Here is a little example that is failing for me:

```js
var fs = require('fs');
var exceljs = require('exceljs');

var stream = fs.createWriteStream('richtTextStream.xlsx');
var wb = new exceljs.stream.xlsx.WorkbookWriter({'stream': stream, 'useStyles': true});
var ws = wb.addWorksheet('rich');
var row = ws.getRow(2);
var cell = row.getCell(2);

cell.value = {
  richText: [
    {
      font: {color: {argb: '99E391'}},
      text: 'This part '
    },
    {
      font: {color: {argb: 'FF6565'}},
      text: ' and that one.'
    }
  ]
};

row.commit();
ws.commit();
wb.commit();
```

If I use a *non* streaming workbook and write the file using `wb.xlsx.writeFile()` then I can see the correct formatting.

Any idea what is going on? Is there some missing rendering logic in the Streaming I/O?

Thanks!

linguamatics-pdenes:
- When streaming, render "richText" values as <c t="inlineStr"><is><r>...</r></is></c>
- Parse inlineStr values correctly, including plain text and rich text

This should fix #409 and hopefully also covers the case in #495.

### DIFF
--- a/lib/xlsx/xform/sheet/cell-xform.js
+++ b/lib/xlsx/xform/sheet/cell-xform.js
@@ -13,7 +13,6 @@ var Enums = require('../../../doc/enums');
 var Range = require('../../../doc/range');
 
 var RichTextXform = require('../strings/rich-text-xform');
-var richTextXForm = new RichTextXform();
 
 function getValueType(v) {
   if ((v === null) || (v === undefined)) {
@@ -47,6 +46,7 @@ function getEffectiveCellType(cell) {
 
 
 var CellXform = module.exports = function() {
+  this.richTextXForm = new RichTextXform();
 };
 
 utils.inherits(CellXform, BaseXform, {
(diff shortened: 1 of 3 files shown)


lib/xlsx/xform/sheet/cell-xform.js
```diff
@@ -251,17 +266,35 @@ utils.inherits(CellXform, BaseXform, {
         this.currentNode = 'v';
         return true;
 
+      case 't':
+        this.currentNode = 't';
+        return true;
+
+      case 'r':
+        this.parser = richTextXForm;//this.map.r;
```
guyonroche: parsing is state-full so this will cause problems if more than one parse is running.

linguamatics-pdenes: Good point! See fix below...

PR #508, opened 2018-02-27
Small README fix

lbguilherme:


guyonroche:
@lbguilherme - thanks for correcting that 👍

PR #549, opened 2018-04-19
Add CSV write buffer support

jloveridge:
Added csv write buffer support. Update types to include `writeBuffer` and modify types so they can be used for both browser and node installations.


jloveridge:
@guyonroche any chance we can get this merged? Are there any changes you want to see before this is mergeable?

PR #562, opened 2018-05-07
List /xl/sharedStrings.xml in [Content_Types].xml only if one of the …

priidikvaikla:
…worksheets use shared strings


priidikvaikla:
@guyonroche What do you think, do you have time to look into it?

guyonroche:
Sorry it's taken so long - I'll publish once tests are good

Issue #488, opened 2018-02-02
Invalid time value Exception


ilijaz:
I got the exception when i opened this [https://yadi.sk/d/9RC1JcPY3S2kHa](url) xlsx document template 
```
null: RangeError: Invalid time value
message: "Invalid time value"
stack: "RangeError: Invalid time value
    at Date.toISOString (<anonymous>)
    at module.exports._format (/var/www/wialon/agri/node_modules/exceljs/dist/es5/xlsx/xform/simple/date-xform.js:17:21)
    at module.exports.render (/var/www/wialon/agri/node_modules/exceljs/dist/es5/xlsx/xform/simple/date-xform.js:36:34)
    at module.exports.render (/var/www/wialon/agri/node_modules/exceljs/dist/es5/xlsx/xform/core/core-xform.js:64:29)
    at module.exports.toXml (/var/www/wialon/agri/node_modules/exceljs/dist/es5/xlsx/xform/base-xform.js:94:10)
    at /var/www/wialon/agri/node_modules/exceljs/dist/es5/xlsx/xlsx.js:441:28
    at new Promise (<anonymous>)
    at new Promish (/var/www/wialon/agri/node_modules/promish/lib/promish-class.js:23:9)
    at module.exports.addCore (/var/www/wialon/agri/node_modules/exceljs/dist/es5/xlsx/xlsx.js:439:12)
    at /var/www/wialon/agri/node_modules/exceljs/dist/es5/xlsx/xlsx.js:608:55
    at <anonymous>"
__proto__: Error {constructor: , name: "RangeError", message: "", …}
```
I did this dirty hack and now everything works fine
```
  this._format = options.format || function (dt) {
    try { return dt.toISOString(); }
    catch(e) { console.log(e); return '' }
  };
```
I don't know why it's happening because i'm newbee in js and don't have enough time to understand exceljs lib at all.

navidshad:
I had the same problem, this error will happen when you use datejs in exceljs options.

optimistex:
How to fix it? Anybody knows?

It is a file that makes the bug:
[xlsx.helper.template.xlsx](https://github.com/guyonroche/exceljs/files/1998345/xlsx.helper.template.xlsx)

optimistex:
My current solution is to make a template through the google documents.

ilijaz:
@guyonroche I've got some skills for that time. I've added a unit test and another PR without catching the exception.
[...]

PR #574, opened 2018-05-24
 Issue #488

dljenkins:
Fix Issue #488 where dt is an invalid time format, catch the exception and return blank string.


bedsDev:
brilliant. it working for me now

PR #578, opened 2018-05-30
Fix some xlsx stream read xlsx not in guaranteed order problem

KMethod:
when stream read some xlsx files (like numbers convert to xlsx),  get values like :
row.values [ ,
  { sharedString: 47 },
  { sharedString: 35 },
  { sharedString: 47 },
  { sharedString: 37 },
  { sharedString: 37 },
  { sharedString: 37 },
  { sharedString: 37 },
  { sharedString: 37 },
  { sharedString: 37 },
]

try to fix this problem

### DIFF
--- a/lib/stream/xlsx/workbook-reader.js
+++ b/lib/stream/xlsx/workbook-reader.js
@@ -21,6 +21,8 @@ var WorkbookPropertiesManager = require('../../xlsx/xform/book/workbook-properti
 var WorksheetReader = require('./worksheet-reader');
 var HyperlinkReader = require('./hyperlink-reader');
 
+var Temp = require('temp');
+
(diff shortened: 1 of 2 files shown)


guyonroche:
@KMethod my concern with this PR is the use of Temp, which won't work in a browser.  The StreamBuf class (in utils) should provide an in-memory alternative

KMethod:
Streaming XLSX are only exposed in NodeJS, so does it matter？

guyonroche:
@KMethod  Sorry this is dragging out - I don't have a huge amount of time these days.
It might matter - this change puts a dependency on the environment (i.e. temp folders).

I'm still not sure what problem this fix is solving.  Can you give more details?

PR #590, opened 2018-06-20
drawing element must be below rowBreaks according to spec or corrupt worksheet

nevace:
I had an issue with adding both images and a pagebreak to a worksheet. Adding them independently was fine but adding them both together result in a corrupt document. 

The issue was related to the ordering of the elements. drawing must be below rowBreaks as seen here:

https://stackoverflow.com/a/12221477/4564456

You can also see the ordering of all the other elements there in case this comes up with new features. Moving drawing and picture below rowBreaks fixes the issue.

I'm currently blocked until this is merged so please let me know if there's anything else I can do to help! :)

Thanks


nevace:
@guyonroche I appreciate you're busy but do you have an ETA for the next release? I'm currently blocked by this issue.

Thanks

guyonroche:
Appologies for the delay - will test and publish today

PR #619, opened 2018-07-30
Add Typescript support for formulas without results

Wolfchin:
It is currently possible, when passing a formula as a value, to omit the result of the formula, and the result will be computed later.
`sheet.getCell("B1").value = { formula: 'SUM(A1:A6)' };`

This PR aims to add this possible syntax to the TypeScript module resolution file.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -333,13 +333,14 @@ export interface CellHyperlinkValue {
 
 export interface CellFormulaValue {
 	formula: string;
-	result?: number | string | Date;
+	result?: number | string | Date | { error: CellErrorValue };
+	date1904: boolean;
 }
 
 export interface CellSharedFormulaValue {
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -333,7 +333,7 @@ export interface CellHyperlinkValue {
 
 export interface CellFormulaValue {
 	formula: string;
-	result: number | string | Date;
+	result?: number | string | Date;
```
Siemienik: Object with property `error`
https://github.com/exceljs/exceljs/blob/master/lib/xlsx/xform/sheet/cell-xform.js#L147

Siemienik: property `date1904`
https://github.com/exceljs/exceljs/blob/master/lib/xlsx/xform/sheet/cell-xform.js#L151

PR #621, opened 2018-07-31
fix(worksheet.state): corrected worksheet state

### LINKED ISSUE #617: If I add more then 1 sheet all sheets are hidden except for the 1st sheet
Hi, I am nor sure if this is a bug or if some new functionality has been added but I am using the latest version of exceljs (1.5.1) and if I add more then one sheet to a workbook then all sheets after the first sheet are hidden and I am not able to make them visible using code.

Here is some sample code to re-produce the issue:

```
const Exceljs = require("exceljs");

let wb = new Exceljs.Workbook();

let sheet1 = wb.addWorksheet("s1");
let sheet2 = wb.addWorksheet("s2");

wb.xlsx.writeFile("myWb.xlsx");

```

If you open this workbook sheet "s2" is hidden. I tried using: 

`sheet2.state = "show";
`
but this doesn't help. Is this a bug or am I missing something?

This works if I use 1.4.6.

Thanks!

mp-ffx:
according to https://github.com/SheetJS/js-xlsx/issues/464 respectively https://msdn.microsoft.com/en-us/library/dd773237%28v=office.12%29.aspx?f=255&MSPPError=-2147217396 and https://msdn.microsoft.com/en-us/library/documentformat.openxml.spreadsheet.sheetstatevalues.aspx the states should/can be 'visible', 'hidden' and 'veryHidden'

this should close #617


Siemienik:
https://github.com/exceljs/exceljs/issues/678

PR #636, opened 2018-08-28
Fix misinterpreted ranges from <definedName>

papandreou:
Hi!

I ran into a file that contained a bunch of formulas and external references in `<definedName>`s, which caused this crash:

```
     Error: Out of bounds. Invalid column letter: OFFSET
      at Object.l2n (lib/utils/col-cache.js:60:13)
      at Object.decodeAddress (lib/utils/col-cache.js:97:24)
      at Object.decodeEx (lib/utils/col-cache.js:189:24)
      at CellMatrix.addCell (lib/utils/cell-matrix.js:19:29)
      at lib/doc/defined-names.js:159:18
      at lib/doc/defined-names.js:157:26
      at module.exports.set model [as model] (lib/doc/defined-names.js:155:11)
      at module.exports.set model [as model] (lib/doc/workbook.js:206:30)
      at lib/xlsx/xlsx.js:348:31
```

Turns out it's caused by what looks like [a rather naive range extraction algorithm](https://github.com/guyonroche/exceljs/blob/acfa8b36ae7d266b36e62891e6c7acf6d228b075/lib/xlsx/xform/book/defined-name-xform.js#L54-L86) that doesn't take into account that the defined names could contain double quoted strings or formulas.

I'm unable to share the original file in its full glory, but I made a new one that reproduces one variant of the error.

I'm really not sure that this is the right fix -- it would probably be better to implement proper tokenization/parsing in `extractRanges` so that the full syntax is understood. Then again, that would be a lot of work, and to be honest I'm have no idea if those ranges are even used for anything important during the parsing?


papandreou:
@guyonroche, is anything holding this back? It happened to us again today, and we'd rather not go back to forking exceljs :)

papandreou:
@guyonroche, sorry to be a ping-in-the-ass :smirk_cat:, but have you had a chance to take a look at this?

PR #666, opened 2018-10-16
Fix incorrect worksheet.state

DaSpawn:
corrects #617:  worksheet.state was incorrect with a value of 'show', requires 'visible'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,7 +236,7 @@ var worksheet = workbook.getWorksheet(1);
 ## Worksheet State
 
 // make worksheet visible
-worksheet.state = 'show';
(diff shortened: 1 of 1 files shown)


felipesilvati:
Great contribution! In addition to that, please emphasize that if you intend to add more than one worksheet, all other worksheets must be set to visible, or you'll get a workbook with only the first worksheet appearing.

adam-beck-sb:
Should be an easy one to get merged in. Thanks @DaSpawn! Unfortunately, I found this fix before figuring it out myself.

Siemienik:
https://github.com/exceljs/exceljs/pull/621

PR #672, opened 2018-10-24
update deps for nodejs 10+

Async0x42:
update all deps


Async0x42:
> package-lock seems to be redundant

Recently I've seen it become more common to ensure the deps match exactly w/ a successful build. If it's not required for this then that's fine.

Siemienik:
@Async0x42 I prefer keep libraries up to date, because when some deps becomes making troubles, it is two possible ways to choose:
1. froze last properly worked version
2. update our code to be up to date

Siemienik:
@Async0x42 Thank you for PL ;)

Async0x42:
> @Async0x42 I prefer keep libraries up to date, because when some deps becomes making troubles, it is two possible ways to choose:
> 
> 1. froze last properly worked version
> 2. update our code to be up to date

It's what I do for my private repos; works for me!

Siemienik:
@Async0x42 have You any idea how fix travis for node v6?

Async0x42:
> @Async0x42 have You any idea how fix travis check for node v6?

Off-hand, I don't to be honest, and I won't have time to dig into it for a little while. Node v6 will be EOL itself in ~2 months (April 2019), would it be OK to just drop support for that a little early?

Ref: https://github.com/nodejs/Release

alubbe:
Should we merge this on 1st of May, when node 6 is no longer supported? :)

Siemienik:
@alubbe @guyonroche , so, v6 is EOL, we should update our Travis.

Siemienik:
https://github.com/exceljs/exceljs/pull/803

package.json
```diff
@@ -70,51 +70,50 @@
...
+    "fast-csv": "^2.4.1",
+    "jszip": "^3.1.5",
+    "moment": "^2.22.2",    
     "unzipper": "^0.9.7",
-    "promish": ">=5.0.2",
-    "sax": "^1.2.2",
-    "tmp": "^0.1.0"
+    "promish": "^5.1.1",
+    "sax": "^1.2.4",
+    "temp": "0.8.3"
```
alubbe: We still want tmp over temp, right?

Siemienik: right,
is it now correct?

PR #679, opened 2018-10-30
Update README.md

hewford:
Please create a demo as to how to send from server to client and download to computer. writefile only writes to the app.


Siemienik:
@hewford please use issues for request like that

PR #695, opened 2018-11-21
Avoid zip parsing when stream gets closed

johnmalkovich100:
If the source stream is dead due some reasons, then no need to continue parsing zip.
Right now in a case of the error, zip is still parsing data even if the source stream is already broken or destroyed which leads to memory leaks and high memory consumption.

We can reproduce this when we interrupt parsing for the big excel files.


alubbe:
Would be awesome to see this land - please let us know @guyonroche if you need any additional info or code to get this merged

johnmalkovich100:
The root problem in the unzip lib

PR #708, opened 2018-12-11
Replaced node-unzip-2 to unzipper package which is more robust

johnmalkovich100:
We are actively using `exceljs` in our company and noticed that sometimes we have a situation when CPU usage and memory consumption are very high if the source stream is broken or dead.

After some investigation I noticed that `node-unzip-2` lib is the reason of this behaviour. Unfortunately it's not well supported and we didn't find the root cause of this problem.

We've found the `unzipper` lib is a well supported fork of the `node-unzip` lib and it's pretty robust. 

In this PR, `node-unzip-2` is replaced to the `unzipper` lib.

Thanks a lot for the cool exceljs lib! 🤝


alubbe:
Would be awesome to see this land. Before this change, we've seen a lot of errors and OOM crashes in production when using the streaming WorkbookReader.

@guyonroche Please let us know if you need any additional info or code to get this merged

alubbe:
@guyonroche any chance that you can take a look at this over the holidays? We have a fork of your repo using node-unzipper and it's working great - but we'd love to return to your npm published package

Siemienik:
@johnmalkovich100 @alubbe Hello, i will invite You to pull request to: https://github.com/p3rio/exceljs
I going to maintain this library there as long it's required ;)

CaioFugii:
unzipper package has no support for *.xls files.

Issue #669, opened 2018-10-19
Font Colors aren't Consistent


huntinghawk1415:
I'm attempting to only change the font color of a row that is being passed into a function:

  ```
write(workbook, sheet, info, row_no, chart_path) {
    const worksheet = workbook.getWorksheet(sheet)
    const row = worksheet.getRow(row_no)
    const no_3 = info[2].slice(0, info[2].length - 1)

    row.font = {color: {argb: 'FFFF0000'}}

    row.eachCell({includeEmpty: true}, (cell, col_no) => {
      col_no === 1 && (cell.value = info[0])
      col_no === 2 && (cell.value = info[1])
      col_no > 3 && no_3.forEach((i, i_no) => {
        col_no - 4 === i_no && (cell.value = i)
      })
    })

    workbook.xlsx.writeFile(chart_path)
      .then(() => console.log('done'))
      .catch(error => console.error(error))
  }
```

Unfortunately when I run this it changes every row between an arbitrary number (i.e. rows 5-14) instead of just the row I specify. Any reason for this that someone can see, or is this a bug?

Siemienik:
@huntinghawk1415 do you know now what was happen?

huntinghawk1415:
@Siemienik Still no idea. The code is a little abrasive to look at to uncover the true cause after a while, so I decided to go with another option for editing the Excel file.

That said, if it helps, the last thing I was able to determine was it was only an issue with already existing Excel files. Files created using exceljs did not exhibit the same behavior and worked as expected.

Siemienik:
@huntinghawk1415 could I please for example file?

huntinghawk1415:
The file I was using contains sensitive data from my company, so I cant share that file. But when I have some time later today or tomorrow I can try to recreate the issue for you with a different file.

Siemienik:
@huntinghawk1415  sure, thanks

Siemienik:
pull request opened
@huntinghawk1415 thanks for details, it was really helpful ;)

iva2k:
Perhaps it's happening due to use of simple assignment in the code of spliceRows(). Javascript assigns objects by reference, so simple assignment just reuses existing object. If later the object properties are changed, they affect all references of the original object. A fix would be a proper clone of the object instead of simple assignment.

julimuchi:
THE SOLUTION: https://github.com/exceljs/exceljs/issues/2055

PR #722, opened 2019-01-14
Properly assigning styles by deep copy

Siemienik:
+ var to const/let refactoring


Siemienik:
#669

lib/doc/row.js
```diff
@@ -367,6 +367,6 @@ Row.prototype = {
     this.hidden = value.hidden;
     this.outlineLevel = value.outlineLevel || 0;
 
-    this.style = value.style || {};
+    this.style = value.style && JSON.parse(JSON.stringify(value.style)) || {};
```
Siemienik: fix is here ;)

alubbe:
Let's do these refactorings as a separate PR - as a matter of fact, this is better done by an automated tool like `prettier` than by hand. Can you simplify this PR? I'll create a separate one for the ES6 refactoring.

lib/doc/row.js
alubbe: I'm not a big fan of this, since there's a couple of subtle ways this can fail or lose things depending on the shape of the object. do we know what values each key of `value.style` can hold? is it just a flat object of `"some string": "another string"` or can it also hold numbers, arrays, functions or other objects?

alubbe:
@guyonroche any chance to get write access to merge PRs like this?

lib/doc/row.js
Siemienik: it is only object that keep [configuration of styles](https://github.com/exceljs/exceljs#styles)
there aren't any functions or getters/setters

Siemienik:
@alubbe I really dislike do changes in all files in one time, it will make a lot of conflicts

alubbe:
on the contrary, doing it one file at a time makes for a really messy git log (making git blame useless) and introducing a highly inconsistent code basis. you are right with regards to the conflicts, so I suggest we merge what we can in the next 14 days - afterwards we merge the big style pr and I update every open PR by hand. how does that sound?

spec/integration/workbook-xlsx-reader.spec.js
```diff
@@ -76,9 +76,9 @@ describe('WorkbookReader', function() {
   });
   
   describe('edit styles in existing file', function(){
-    beforeEach(async function(){
+    beforeEach(function(){
```
alubbe: why? I quite like moving things to async/await. the test suite runs on node, so there's no transpiling here

lib/doc/row.js
alubbe: gotcha. Can you help me understand the performance impact here?
`JSON.parse(JSON.stringify())` is quite slow and it would suck if this actually runs for all (or most) rows. Also, it increases the memory overhead a bit. How many rows will have this property?

spec/integration/workbook-xlsx-reader.spec.js
Siemienik: we currently support node v6 :(
https://travis-ci.org/exceljs/exceljs/builds/479501545?utm_source=github_status&utm_medium=notification

alubbe: damn, right you are :(

alubbe:
@Siemienik could you help understand if the object cloning is happening on the hot path for each row? how often do rows have this style property?

Siemienik:
@alubbe as offen as style was set. in #669 is really well described example when it works wrong.

probably this solution isn't best of all, but it's best what i found.

lib/doc/row.js
Siemienik: personally I think it's much better to write style cloning function in utils ;)

alubbe:
Let's merge it then - I still don't have access to do it though :(

lib/doc/row.js
BridgeAR: Using JSON is indeed pretty slow to copy things. You could use e.g., https://github.com/davidmarkclements/rfdc for now (I would normally just recommend to use object spread but that's not available in versions below 8 and slow before v.10 or v.11 :smile:).

@alubbe the current API does not seem to have performance in mind. If the user changes any input that should be registered in a clean way so all internals can work side effect free. This does not seem to be possible without the penalty here. Using getters and setters might reduce the overhead by only copying values that have to be changed but as far as I see it that would probably be pretty brittle with the current code.

alubbe: I agree - let's merge it and then actually benchmark why this library is slow and fix those, as opposed to pre-optimizing a bugfix

PR #736, opened 2019-02-02
add Worksheet.state typescript definition fix #714

### LINKED ISSUE #714: Worksheet State does not exist in index.d.ts
Due to this reason it is not possible to make visible newly added worksheets.
Following gives errors when used in Agular 7: 
1)            var worksheet = workbook.addWorksheet('Title');
             worksheet.state='visible'

2)  var worksheet = workbook.addWorksheet('Title',{ state: 'visible' });

ilyes-kechidi:


index.d.ts
```diff
@@ -848,6 +848,11 @@ export interface Worksheet {
 	 */
 	pageSetup: Partial<PageSetup>;
 
+	/**
+	 * Worksheet State
+	 */
+	state:  'show' | 'visible' | 'hidden' | 'veryHidden';
```
alubbe: minor nitpick: please use only one space after `:`

alubbe:
I checked that the default value of the work sheet state is `show`, but from my understanding the same holds true for the options object when adding worksheets since it's just passed down to the constructor (in this PR, it's default value is `visible`). In that case, we should define this enum once and the re-use it in both type definitions. what do you think @Siemienik @guyonroche ?

Siemienik:
@alubbe fully agree, 

as far as i know, 'show' is invalid value, it should be 'visible' 
#621

Issue #697, opened 2018-11-22
How to insert data between two rows of an existing excel


chaonsf:

chaonsf:
A blank face

chaonsf:
spliceRows  
But the next row has completely changed its style.

Issue #305, opened 2017-04-20
After SpliceRows, shifted rows lost cell names


supermyz:
Hi
Thanks for this great lib, helps a lot, here is the issue: the shifted rows lost information, not just the names but also their styles

`
var workbook = new Excel.Workbook();
workbook.xlsx.readFile(filename).then(function(x) {
	let worksheet = workbook.getWorksheet(1);
	worksheet.spliceRows(5, 1);
	workbook.xlsx.writeFile(filename).then(function() {
		console.log('file saved');
	});
}).catch(function(err) {
	console.log(err);
});

`

FluffyNevyn:
I am also seeing this issue - glad I checked the open issues before submitting a new one.  After a call to worksheet.spliceRows(...) any rows that get Moved lose all style information (font, fill, border, etc...)

iva2k:
Should this issue be closed? (PR with a fix was merged)

nhd2106:
> I am also seeing this issue - glad I checked the open issues before submitting a new one. After a call to worksheet.spliceRows(...) any rows that get Moved lose all style information (font, fill, border, etc...)

i still got this issue, i'm using lastest version

baelene:
I have also the issue with merged cells: after spliceRows, the merged cells below the deleted rows are not (merged) anymore.

tsensei:
@baelene Did you get any solution for this ?

PR #737, opened 2019-02-03
Fix existing row styles when using spliceRows

cxam:
This PR fixes the style issues reported in https://github.com/exceljs/exceljs/issues/305 and https://github.com/exceljs/exceljs/issues/697 to copy over any existing styles already applied to rows. I have also included additional tests for this fix.

### DIFF
--- a/lib/doc/worksheet.js
+++ b/lib/doc/worksheet.js
@@ -330,6 +330,10 @@ Worksheet.prototype = {
         rSrc = this._rows[i - 1];
         if (rSrc) {
           this.getRow(i + nExpand).values = rSrc.values;
+          // eslint-disable-next-line no-loop-func
+          rSrc.eachCell({ includeEmpty: true }, function(cell, colNumber) {
+            this.getRow(i + nExpand).getCell(colNumber).style = cell.style;
+          }.bind(this));
           this._rows[i - 1] = undefined;
         } else {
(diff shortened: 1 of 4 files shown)


alubbe:
I'm really not familiar with how styles work, but is this related to https://github.com/exceljs/exceljs/pull/722 @Siemienik ?

Siemienik:
@alubbe it is quite different issue

#722 prevent about duplicating references when reading.

lib/doc/worksheet.js
```diff
@@ -330,6 +330,10 @@ Worksheet.prototype = {
         rSrc = this._rows[i - 1];
         if (rSrc) {
           this.getRow(i + nExpand).values = rSrc.values;
+          // eslint-disable-next-line no-loop-func
+          rSrc.eachCell({ includeEmpty: true }, function(cell, colNumber) {
+            this.getRow(i + nExpand).getCell(colNumber).style = cell.style;
```
Siemienik: hmm.. have we 100% that `cell.style` object hasn't multiple references from some different cells? 
may should we use deep copy here?

@cxam @alubbe what are You thinking about it?

cxam: @Siemienik do you think that would be necessary because the row that we get the style for each cell gets set to undefined anyway?

Siemienik: I worry about case like this one #669

Issue #738, opened 2019-02-03
Keep original image dimensions


kticka:
How to keep original image dimensions instead stretching it through cells? Why top-left position isn't enough when inserting image?

Siemienik:
#702

PR #702, opened 2018-12-01
Improvements for images (correct reading/writing possitions)

Siemienik:
Improvements for image`s positioning/stretching and their properly writing to xlsx and readign from xlsx with fully backward compatibility.

### Before that changes, 

Reading images positions works **wrong**. This example show it very well,
original file looks that:
image
after i open this file and save **without any changes**:
image

## Backward compability

My solution was designed to handle current `col`/`row` and calculate it to `nativeCol`, `nativeColOff`, `nativeRow`, `nativeRowOff`. A lot of tests was written to ensure properly working with dependents apps .

I will be glad for quick checking/approval this pull request, I am open for your comments and ready to make some fixes.

### issues
https://github.com/guyonroche/exceljs/issues/650
https://github.com/guyonroche/exceljs/issues/467

### Use now
if You need it quicly, i have build branch in my repo:
https://github.com/p3rio/exceljs/tree/build

npm:
```
npm install github:p3rio/exceljs#build --save
```
or add to package.json:
```
    "dependencies": {
        "exceljs": "github:p3rio/exceljs#build"
    }
```

Issue #650, opened 2018-09-17
Image location don't respect Column width


yuvpf:
I set custom column width while defining sheet columns. When adding an image using coordinate (50% column to use), the image don't respect my custom width. 

**Code :**

      sheet.columns = [
          { header: 'Id', key: 'id', width: 10 },
          { header: 'Name', key: 'name', width: 32 },
      ];

      var imageId2 = workbook.addImage({
        base64: myBase64Image,
        extension: 'png',
      });

      sheet.getColumn(4).width = 38;
      // sheet.addImage(imageId2, 'B2:D6');
      sheet.addImage(imageId2, {
                          tl: { col: 1, row: 1 },
                          br: { col: 1.5, row: 4 }
                        });

      sheet.addImage(imageId2, {
                          tl: { col: 3, row: 1 },
                          br: { col: 3.5, row: 4 }
                        });

      workbook.xlsx.writeBuffer({
                  base64: true
              })
      .then(function(xls64) {
          // done
          var data = new Blob([xls64], { type: "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet" });
          saveAs(data, "output.xlsx");
      });

**Actual result :** 

![image](https://user-images.githubusercontent.com/39988720/45623646-a7a86000-baa5-11e8-9e94-2d4d00fd6a9b.png)

**Expected result:**

![image](https://user-images.githubusercontent.com/39988720/45623683-c4dd2e80-baa5-11e8-8c9f-c9f6aca135ae.png)

GendelfLugansk:
@yuvpf did you find any workaround?

Siemienik:
@yuvpf @GendelfLugansk 
Hello guys, i fix it in pull request #702, 
if You need it quicly, i have build branch in my repo: 
https://github.com/p3rio/exceljs/tree/build

npm:
```
    "dependencies": {
        "exceljs": "github:p3rio/exceljs#build"
    }
```

ShraddhaJ11:
Hi @Siemienik  Still facing image alignment issue with custom column width.

Siemienik:
hmm.. it works for me really well, which version have you?

ShraddhaJ11:
> hmm.. it works for me really well, which version have you?

I am using 4.0.1. How you are calculating colOff and rowOff?

Siemienik:
it's enough to specify `col` and `row` values. collOff and rowOff will be calculated during saving

ShraddhaJ11:
[...]

PR #746, opened 2019-02-14
Add support for adding images anchored to one cell

karlvr:
Addresses #738 #650 and is related to #702.

There is a bug in `cell-position-xform.js` that assumes that columns are 640,000 EMU wide and 180,000 EMU tall. This assumption is broken if you customise cell width or height, and the bug appears if you use fractional columns or rows to try to position your image.

This PR adds support for adding images using just a `tl` and then providing an `ext` that contains a `width` and `height` in pixels. The width and height are converted internally to EMU using a conversion at 96dpi. That _seems_ to be the right thing to do, but something I'm a little unsure about!

This change _should_ read as well as write the one-cell-anchor-xform, although I have only used writing so far but there is a test for reading and writing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1225,6 +1225,17 @@ ws.addImage(imageId, {
 });
 ```
 
+### Add image to a cell
+
+You can add an image to a cell and then define its width and height in pixels at 96dpi.
+
+```javascript
+worksheet.addImage(imageId2, {
+  tl: { col: 0, row: 0 },
+  ext: { width: 500, height: 200 }
+});
+```
+
 ## File I/O
 
 ### XLSX
--- a/index.d.ts
+++ b/index.d.ts
@@ -770,6 +770,11 @@ export interface ImageRange {
 	br: { col: number; row: number };
 }
 
+export interface ImagePosition {
+	tl: { col: number; row: number };
+	ext: { width: number; height: number };
+}
+
 export interface Range extends Location {
 	sheetName: string;
 
@@ -1022,7 +1027,7 @@ export interface Worksheet {
 	 * Using the image id from `Workbook.addImage`,
 	 * embed an image within the worksheet to cover a range
 	 */
-	addImage(imageId: number, range: string | { editAs?: string; } & ImageRange): void;
+	addImage(imageId: number, range: string | { editAs?: string; } & ImageRange | { editAs?: string; } & ImagePosition): void;
 
 	getImages(): Array<{
 		type: 'image',
--- a/lib/xlsx/xform/drawing/drawing-xform.js
+++ b/lib/xlsx/xform/drawing/drawing-xform.js
@@ -11,12 +11,18 @@ var XmlStream = require('../../../utils/xml-stream');
 
 var BaseXform = require('../base-xform');
 var TwoCellAnchorXform = require('./two-cell-anchor-xform');
+var OneCellAnchorXform = require('./one-cell-anchor-xform');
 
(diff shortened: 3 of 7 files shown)


karlvr:
You can test it by adding `https://github.com/karlvr/exceljs#integration` to your `package.json`.

WebDeg-Brian:
I just added it to my `package.json` and works like a charm. Is `ext` compulsory?

karlvr:
I think you need to provide either `ext` or use the existing method. I guess we could allow `ext` to be omitted if we just used the image width and height?

WebDeg-Brian:
@karlvr definitely. If `ext` is not specified, we should use the original image's dimensions. One might get the image programmatically somewhere without knowing the exact width and height. Obviously there are libraries out there that help with finding images' sizes, but still would be nice if your PR does it out of the box

karlvr:
It sounds reasonable. I wonder if there is other image loading code in this library. Might well be!

guyonroche:
Looking into this now. Looks good, my only comment so far would be to offset tl by -1, -1 like the two-cell anchor. I'll look into this

lukasz-t-michalski:
Hi @karlvr , I have an issue using this functionality - exporting images looks great but they aren't getting deleted when I deleted a row from the Excel file. When I use use the default (tl: br:) the attribute editAs: 'oneCell' resolved this issue, but it doesn't seem to work when I use tl: ext: Did anyone else came across this problem?

Devil7DK:
@karlvr I have issues when using ext for image size.

The excel says that "**We found unreadable content in 'Export.xlsx'. Do you want to recovery the contents of this workbook?**" After repairing the file using excel, its opening fine.

The error log says "**Repaired Records: Drawing from /xl/drawings/drawing1.xml part (Drawing shape)**". So i compared the drawing1.xml from both repaired and unrepaired workbooks. And found that if i remove **editAs="oneCell"** attribute from all **xdr:oneCellAnchor** tags, the error is gone.

censys-git:
If you set ext with height and width it works but you get the MS error and repair. I did a diff on the drawings xml of the original and repaired. No changes what-so-ever were made other than on the root xdr:oneCellAnchor where excelJS sets an attribute 'editAs="absolute"' or 'editAs="oneCell"', Microsoft wipes that out (no attribute at all). That is the only change. So very confused on what the heck we need to do to make this work. Should you not be setting this attribute? Is it defined elsewhere where MS is barfing?

YSHIDM:
> 如果您使用高度和宽度设置ext，它可以工作，但会出现MS错误并修复。我对原件的图纸xml进行了比较并修复了。除了在根xdr：oneCellAnchor上之外，没有进行任何更改，其中excelJS设置了一个属性“editAs=”absolute“或”editAs=“oneCell”“，Microsoft将其清除（根本没有属性）。这是唯一的变化。所以非常困惑我们需要做什么才能完成这项工作。您不应该设置此属性吗？它是否在 MS 所在的其他地方定义？

+1

YSHIDM:
> If you set ext with height and width it works but you get the MS error and repair. I did a diff on the drawings xml of the original and repaired. No changes what-so-ever were made other than on the root xdr:oneCellAnchor where excelJS sets an attribute 'editAs="absolute"' or 'editAs="oneCell"', Microsoft wipes that out (no attribute at all). That is the only change. So very confused on what the heck we need to do to make this work. Should you not be setting this attribute? Is it defined elsewhere where MS is barfing?

+1

PR #799, opened 2019-04-30
Add support for repeated columns on every page when printing.

FreakenK:


alubbe:
PR and tests look good, but I have never used this feature of exceljs (or excel). Can someone review it who is more closely aware of how this is used?

PR #806, opened 2019-05-06
load method definition added to index.d.ts, browser section added to package.json, resolves #804

### LINKED ISSUE #804: Impossible to use browser version in TypeScript application
There is some sort of a workaround that allows to import browser build of exceljs
`const Excel = require('exceljs/dist/es5/exceljs.browser');`
but it does not work in a typescript application.
I've solved it for now by changing `"main": "./dist/es5/index.js",` to `"main": "./dist/es5/exceljs.browser.js",` in package.json in fork. In addition to it I had to add `load` method definition to `index.d.ts`. It works, but it is an ugly temporary solution.

I will try to solve this issue without breaking changes and contribute the solution if I succeed.

Pr05p3ro:


Pr05p3ro:
Um... It looks like master build is failing? o_O
https://travis-ci.org/exceljs/exceljs/builds/528873143

package.json
```diff
@@ -117,6 +117,9 @@
   },
   "main": "./dist/es5/index.js",
   "types": "./index.d.ts",
+  "browser": {
```
alubbe: Can't we just use `"browser": "./dist/es5/exceljs.browser.js"`?

Pr05p3ro: Sure, it is possible (already changed).

PR #815, opened 2019-05-14
Do not use a promise polyfill on modern setups

alubbe:
To make the most of async/await (performance & stack traces), we need to use the native Promises. This PR does not change that the default export uses `promish`, but it does add two new exports that allow access to the modern codebase and native promises. I've also fixed a couple of eslint failures in the affected files.

```javascript
const Excel = require('exceljs/modern.nodejs');
import Excel from 'exceljs/modern.browser';
```

### DIFF
--- a/lib/csv/csv.js
+++ b/lib/csv/csv.js
@@ -3,7 +3,7 @@
 const fs = require('fs');
 const csv = require('fast-csv');
 const moment = require('moment');
-const PromishLib = require('../utils/promish');
+const PromiseLib = require('../utils/promise');
 const StreamBuf = require('../utils/stream-buf');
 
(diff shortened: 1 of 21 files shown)


lib/csv/csv.js
```diff
@@ -121,7 +121,7 @@ CSV.prototype = {
             }
             if (value instanceof Date) {
               if (dateFormat) {
-                dateUTC ? moment.utc(value).format(dateFormat) : moment(value).format(dateFormat);
+                return dateUTC ? moment.utc(value).format(dateFormat) : moment(value).format(dateFormat);
```
alubbe: I'm pretty sure the missing return here was a bug

Issue #705, opened 2018-12-05
Corrupted XLSX output due to invalid sheet name


loreanvictor:
So I noticed that while some softwares could open generated XLSX files using `exceljs` (like Numbers on MacOS and Google Sheets), Microsoft Excel (or Excel Online) was unable to open the files, giving me an error indicating that the file is corrupted. After simplifying the code and isolating the issue, I realized that for some reason MS Excel assumes a maximum length of 31 characters for Worksheet titles and as a result was deeming the generated file corrupt.

So basically, this snippet works fine:

```javascript
const excel = require('exceljs');

let book = new excel.Workbook();
let sheet = book.addWorksheet('random sheet name that is valid');

sheet.columns = [
  {header: 'First Name', key: 'firstName'},
  {header: 'Last Name', key: 'lastName'}
];

sheet.addRow({ firstName: 'John', lastName: 'Wick' });
sheet.addRow({ firstName: 'Elliot', lastName: 'Alderson'});

book.xlsx.writeFile('test.xlsx');
```

While this one doesn't:

```javascript
const excel = require('exceljs');

let book = new excel.Workbook();
let sheet = book.addWorksheet('random sheet name that is not valid');

sheet.columns = [
  {header: 'First Name', key: 'firstName'},
  {header: 'Last Name', key: 'lastName'}
];

sheet.addRow({ firstName: 'John', lastName: 'Wick' });
sheet.addRow({ firstName: 'Elliot', lastName: 'Alderson'});

book.xlsx.writeFile('test.xlsx');
```

Since MS Excel is what most end-users will use to open automatically generated files by a web-service, I would highly recommend adding an error/warning message for when this case occurs, a toggle-able validation that is on by default, or at least a hint to the documentation.

Siemienik:
@loreanvictor you are right, 
maybe shall You write pull request with this validation? :)

omurbekjk:
@Siemienik Any ideas how to validate correctness of excel? maybe I will try to create pull request.

Siemienik:
@omurbekjk somethink like that works for me:
```javascript
       name = name.replace(/[\\\/*\[\]?]/g, '.');  //replace forbidden chars
        if (!scope.isUnique(name))) { //name has to be unique
            name = scope.next(name);
        }        
        name = name.length > 31 ? name.substr(name.length - 31) : name; //max 31 letters

```

Issue #664, opened 2018-10-12
Defined Names Break/Corrupt Excel File into Repair Mode


huntinghawk1415:
### *Problem*

After a bit of time trying to figure out why my files were always being corrupted when using `workbook.xlsx.writeFile()`, I came to find out it was the named ranges (under the formulas tab) in the template I was reading that was causing the issue - more specifically, it was the way the ranges were being extracted and re-entered into a "new" xlsx file. To give you an idea of what I was going through, here's the xml for causing the issue:

**First, the Excel Error**

```
Excel found unreadable content in 'test.xlsx'. Do you want to recover the contents of this 
workbook? If you trust the source of this workbook, click Yes
```

**Working XLSX Example**
```
<definedName name="_xlnm.Print_Area" localSheetId="0">'12pm'!$C$1:$AX$12</definedName>
<definedName name="_xlnm.Print_Titles" localSheetId="0">'12pm'!$C:$E</definedName>
```

**How ExcelJS is Rewriting it**
```
<definedName name="_xlnm.Print_Area" localSheetId="0">12pm!C1:AX12</definedName>
<definedName name="_xlnm.Print_Titles" localSheetId="0">&apos;12pm&apos;!$$C:$$E</definedName>
```

As you can see, the first definedName named `_xlnm.Print_Area` had all of its apostrophes and dollar signs removed. `_xlnm.Print_Titles` however is replacing every apostrophe with `&apos;` entities, and for some reason, it is adding a new dollar sign on every run of the program (i.e. if I were to run it again, it would rewrite it as `&apos;12pm&apos;!$$$C:$$$E`)

Fortunately, I found out the apostrophes and dollar signs aren't necessary for these kind of simple names, so the `_xlnm.Print_Area` name works perfectly fine. The unfortunate side of things is the `_xlnm.Print_Titles` corrupts the file and sends excel into repair mode, at which point it manually deletes the names causing the issue.

Because I got very impatient trying to navigate through the code to find out why the latter of the two was doing all sorts of weird stuff (hint hint cough cough document the code a little more thoroughly please ^^), I decided to hard-patch the issue with a little filter.

### *Solution*

If you're having the same issue as I am, (v. 1.6.1) navigate to the `exceljs/dist/es5/xlsx/xform/book/defined-name-xform.js` file, go to line 25:25, and replace the

`xmlStream.writeText(model.ranges.join(','));`

with

[...]

PR #822, opened 2019-05-18
fix issue with print area defined name corrupting file

donaldsonjulia:
PR includes:

- Fix file corruption caused by `printArea` defined name. As described in issue below:
https://github.com/exceljs/exceljs/issues/664

- Lint updates required by prettier in committed files.

- Handle formatting of reconciled printArea, printTitlesRow, and printTitlesColumn the same, so all are serialized as `A:A` (not `$A:$A`), in following with the documented format for specifying a sheet's printTitlesRow, etc. ( `ws.pageSetup.printTitlesRow = 'A:B';` )

- Update tests accordingly (per format changes above) and add test `testPrintAreaOut.js`


donaldsonjulia:
In my project I experienced the corruption as a result of adding a printArea to a sheet with a name with a space in it ("sheet title", for example). That is the main issue I observed that motivated this PR.

donaldsonjulia:
@Siemienik This PR precedes my PR for headerFooter, in order to separate concerns.

spec/.eslintrc
```diff
@@ -7,7 +7,8 @@
     "array-bracket-spacing": ["off"],
     "no-sparse-arrays": ["off"],
     "object-property-newline": ["off"],
-    "no-underscore-dangle": ["off"]
+    "no-underscore-dangle": ["off"],
+    "no-unused-expressions": ["off"]
```
guyonroche: I think it is better to leave no-unused-expressions turned on. We have configured dirty-chai which forces all the assertion 'properties' into functions. So expect(blah).to.be.undefined needs to be expect(blah).to.be.undefined(). This is to avoid common errors like expect(blah).to.be.defined - which does absolutely nothing but I have seen it often enough!

guyonroche:
@donaldsonjulia - Thank you for contributing :-) Are you sure this fixes issue #705 ? I thought that issue was about worksheet names

spec/.eslintrc
donaldsonjulia: Noted - I have removed this and updated test with `to.be.undefined()`

donaldsonjulia:
@guyonroche You are correct about issue #705, sorry for the confusion! I removed it from the description. Late night PR summaries with too many tabs in my browser :)

PR #824, opened 2019-05-21
Fix reading of a file with missing entries in the zip container

papandreou:
Stumbled upon some errors when someone uploaded an .xlsx file which had a wrong directory structure in the zip container, so some parts were ignored, which meant that they weren't available when the parsers expected it.

This prevents exceljs from breaking in those cases.


papandreou:
Haha, ironically it seems like I forgot to add the "missing bits" fixture file. Will push it tomorrow.

PR #829, opened 2019-05-22
Introduce async/await 

alubbe:
Here it comes! 🎉 

In this PR, I've not modified any tests so that we know that these changes are a pure refactoring.
Unfortunately, `async` uses the internal `Promise` object and we cannot offer our users to use a different Promise library. To be backwards compatible with promish, I've added babel polyfill, but I would expect that most consumers of this library will start to use the modern entrypoints.

Looking forward to your feedback!

### DIFF
--- a/spec/browser/exceljs.spec.js
+++ b/spec/browser/exceljs.spec.js
@@ -1,7 +1,7 @@
-'use strict';
+/* global ExcelJS */
+// ExcelJS is a global injected by `./dist/exceljs.js` during jasmine's setup
 
-require('babel-polyfill');
-const ExcelJS = require('../../lib/exceljs.browser');
+'use strict';
 
 function unexpectedError(done) {
   return function(error) {
--- a/spec/browser/exceljs.spec.js
+++ b/spec/browser/exceljs.spec.js
@@ -1,6 +1,6 @@
 'use strict';
 
-require('babel-polyfill');
+require('@babel/polyfill');
 const ExcelJS = require('../../lib/exceljs.browser');
 
 function unexpectedError(done) {
--- a/spec/browser/exceljs.spec.js
+++ b/spec/browser/exceljs.spec.js
@@ -1,6 +1,6 @@
 'use strict';
 
-require('@babel/polyfill');
+require('babel-polyfill');
 const ExcelJS = require('../../lib/exceljs.browser');
 
 function unexpectedError(done) {
--- a/.eslintrc
+++ b/.eslintrc
@@ -32,7 +32,7 @@
     "prefer-rest-params": ["off"],
     "quotes": ["error", "single"],
     "semi": ["error", "always"],
-    "space-before-function-paren": ["error", "never"],
+    "space-before-function-paren": ["error", {"anonymous": "never", "named": "never", "asyncArrow": "always"}],
     "strict": ["off"],
     "class-methods-use-this": ["off"],
     "no-return-assign": ["off"],
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,6 @@ package-lock.json
 
 # optional switch off tests
 .disable-test-*
+
+# jasmine
+_SpecRunner.html
--- a/.prettier
+++ b/.prettier
@@ -1,5 +1,6 @@
 {
   "bracketSpacing": false,
   "printWidth": 160,
-  "trailingComma": "all"
+  "trailingComma": "all",
+  "bracketSpacing": false
 }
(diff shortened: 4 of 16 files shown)


spec/browser/exceljs.spec.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-require('babel-polyfill');
+require('@babel/polyfill');
```
Siemienik: un luckily, this break backward compatibility :(

Siemienik: unluckily, this break backward compatibility :(

guyonroche: I'm not hard against breaking changes - we can bump major version for that.
I've never been particularly happy with the PromiseLib compromise - what happened was, the original promise lib was Bluebird, which was ok, but when I discovered how many megabytes it used it really hurt the browser bundle, that's when Promish was introduced. But the switch broke some clients so the option to inject the Promise lib of choice was added.
Now that native promise supports "finally" (as of Node v10.3) it might be time to fix it properly.
Since it would be a breaking change, we should announce it properly, give users time to adjust, etc

alubbe: Fun fact: we don't even need `Promise.finally` anymore, since with async/await it's just `try/catch/finally`.

@guyonroche I think it's not that big a change, even our tests are passing without any real modifications to them. I think this change doesn't need a huge amount of forewarning, unless there's other stuff that you'd like to clean up for exceljs v2?

Siemienik:
@alubbe hmm... What are we going to do with this pull request? I think it's a great idea to merge it but there is a lot of conflicts currently

alubbe:
I've rebased the PR, fixed all conflicts and reverted to using the previous version of babel-polyfill, so that we're fully backwards compatible. Let me know what you guys think, I'd love to merge it! (And then we can start refactoring the tests to use async/await)

alubbe:
Also, as I reported via my issue, travis no longer runs any tests - does anyone know why?
You'll have to check out this branch locally and run the tests there to verify that everything is green (it is on my machine)

alubbe:
Tests are green: https://travis-ci.org/exceljs/exceljs/builds/574842008

spec/browser/exceljs.spec.js
alubbe: I've reverted this change, this PR now uses the version of babel-polyfill that is currently on `master`

alubbe:
I'd be fine with bumping the version, but in that case let's remove a lot of old leftovers:
- Bump @babel-polyfill version
- Replace the current entry with their ES6 version
- ... other stuff?

So in short, it would be great to see this PR merged, and then let's spend a little time cleaning up more stuff, and then bump the version.
So maybe let's merge the TypeScript PRs, bump v1 and then merge this PR?

alubbe:
Do we have anyone who feels comfortable merging the TS PRs? To me that's the only thing outstanding to merge this PR and get started on the next major version.

guyonroche:
@alubbe I've merged this and all the TS PRs I could see. Are there any I've missed?
I'd like to clean up the whole PromiseLib mess before bumping the version then we can move forwards from there :-)

guyonroche:
@alubbe @Siemienik @coldhiber 
Hi Guys - I'm thinking it might be time for yet another major change - this time to refactor the published artefacts...

The Main NPM artefact (what you get when you require/import 'exceljs' would come from the lib folder (we would probably want to bump engines directive to ">=8")

To maintain support for those that still need ES5 in node we will babel into dist/exceljs.es5.js and dist/exceljs.es5.min.js

Does anyone still use browserify ? If so we can keep them too but my thoughts are that modern web devs using react and angular (et al) can use the node type imports

What do you think?

alubbe:
Yes please! This was the main reason I introduced the `modern` files and I agree, they should be the default now. 

Keeping the es5 build is nice, but we might also be fine completely dropping it, I don't think that developers that manipulate excel files via JS targeting old browsers will not be aware of transpilation.

guyonroche:
@alubbe I've done the refactor and tidied up all of the tests and pushed to master
"verquire" in the specs is promoted to a global now and will switch on the EXCEL_BUILD env to pull in lib/ or dist/es5/ code.

I think keeping the es5 still has value - it's still likely that there are builds out there where selectively including dependencies in the transpilation might be problematic.

I feel the same is true with the browserify export. I know modern web dev is now overwhelmingy react, angular (and similar) and looking at google trends comparing webpack with browserify mirrors this, but once again - there will be builds out there still relying on the browserify export.

Given that there's practically next to no cost in keeping them, I think we might as well

I'm going to leave these changes in master for a bit, I'd like to test it out as a github dependency to check it's all ok...

alubbe:
Just saw that you released 3.0.0, looks great - let's wait for feedback from usage in the real world ;)

Issue #398, opened 2017-09-11
Limit worksheet name length to 31 characters


vid:
Excel will show a warning if worksheet names are longer than 31 characters. To maintain full compatibility, Exceljs should limit worksheet lengths, or have some kind of warning.

guyonroche:
Both warning and truncating names to 31 chars deployed in v1.12.1

PR #831, opened 2019-05-23
Only keep at most 31 characters for sheetname

kaleo211:
Addressing [Issue 398](https://github.com/exceljs/exceljs/issues/398).

### DIFF
--- a/lib/doc/workbook.js
+++ b/lib/doc/workbook.js
@@ -44,7 +44,11 @@ class Workbook {
 
   addWorksheet(name, options) {
     const id = this.nextId;
-    name = name || `sheet${id}`;
+
+    name = (name || `sheet${id}`).substring(0, 31);
+    if (this._worksheets.find(ws => ws && ws.name === name)) {
+      throw new Error(`Worksheet name already exists: ${name}`);
+    }
 
     // if options is a color, call it tabColor (and signal deprecated message)
     if (options) {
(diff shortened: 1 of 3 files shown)


alubbe:
Could you please add 2 tests for this:
1. a 31 character name is unchanged
2. a 32 character name is truncated

alubbe:
@guyonroche I think we could print a warning when this happens, or maybe even throw an error, asking the user of the library to only submit short names. What do you think?

Siemienik:
> Could you please add 2 tests for this:
> 
> 1. a 31 character name is unchanged
> 2. a 32 character name is truncated
 
3. two names with more characters than 31, where first 31  characters are equals

eq.
```
ThisIsAnExampleNameWith31Chars_A
ThisIsAnExampleNameWith31Chars_B
```

4. similar but with first letter different:
```
```
A_ThisIsAnExampleNameWith31Chars
B_ThisIsAnExampleNameWith31Chars
```

Because as far as I know, names can't be equals

kaleo211:
> A_ThisIsAnExampleNameWith32Chars

Thank you. I broke down those tests, hope this also works.

guyonroche:
@alubbe 
> @guyonroche I think we could print a warning when this happens, or maybe even throw an error, asking the user of the library to only submit short names. What do you think?
I don't think throwing an error is necessary - but something like a console warning would be appropriate.
This looks good - will merge soon

lirantal:
@guyonroche just chiming in here after I discovered now that this broke some things for me. In the future, I would consider changes like this one as a breaking change. The problem it introduced is that because it is truncating to 31 characters there is now increased potential for duplicate names which will trigger an error thrown.

Regardless, keep up the great work on this library 👏

Issue #696, opened 2018-11-22
Default row height doesn't work ?


chaonsf:
ws1.properties.defaultRowHeight=30;  
but no change in height

Issue #634, opened 2018-08-28
The worksheet.properties.defaultRowHeight  can't work!! How to set the rows height, help!!


GO-DIE:
And I found all the way  of set row height are no work, why?

yilikun:
type you code

KingRial:
The same problem is discussed here: https://github.com/guyonroche/exceljs/issues/422

Issue #422, opened 2017-10-16
row height doesn't apply to row


qiandongyq:
Default row height does not apply to the excel generated.
```
 let infoSheet = this.workbook.addWorksheet("Info", {properties:{ defaultRowHeight: 30 }});
```
or
```
infoSheet.properties.defaultRowHeight = 30;
```
![row_height](https://user-images.githubusercontent.com/24208956/31632907-9b463368-b28c-11e7-895b-c766a51bcf6b.jpg)

gelodgreat:
I use this and it works.
sheet.properties.defaultRowHeight = 40;

lnmput:
I this
`sheet.properties.defaultRowHeight = 40;`
not work

JoeyAspire:
I had the same issue, but i find the following code works:
sheet.getRow(2).height = 100;
sheet.getRow(3).height = 200;
...

chenvan:
the same problem, just want to add a min height style.

sohail85:
Having same issue

ScottPeterJohnson:
For other people having this issue: Are you first opening the generated excel files on the Windows version of Excel 2016?
I've seen something on that platform where Excel will do weird resizing of rows on a worksheet. I painstakingly tracked this down to exceljs not generating a <sheetView/> element, causing Excel to flip. I fixed this by adding an empty view to all of my worksheets via `worksheet.views = [{}]`

yuanoook:
@ScottPeterJohnson  thanks very much.

`worksheet.views = worksheet.views || [{}]` works for me

byeokim:
I experienced the same "weird resizing of rows". (FWIW I had first opened the exceljs-generated `.xlsx` file on the macOS version of MS Excel) And I found that MS Excel-generated `.xlsx` file has the following XML element in `xl/worksheets/sheet1.xml` and @ScottPeterJohnson's `worksheet.views = [{}]` also adds the element:

```xml
...

<!-- <dimension ... /> -->
<sheetViews>
  <sheetView workbookViewId="0" />
</sheetViews>
<!-- <sheetFormatPr ... /> -->

...
```

This addition seems to solve the issue but I couldn't find any explanation from [MS Documentation](https://docs.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.sheetview?view=openxml-2.8.1):

[...]

PR #855, opened 2019-06-11
Fix defaultRowHeight not working

autukill:
https://github.com/exceljs/exceljs/issues/696

https://github.com/exceljs/exceljs/issues/634

https://github.com/exceljs/exceljs/issues/422


lib/xlsx/xform/sheet/sheet-format-properties-xform.js
```diff
@@ -13,6 +13,7 @@ class SheetFormatPropertiesXform extends BaseXform {
         outlineLevelRow: model.outlineLevelRow,
         outlineLevelCol: model.outlineLevelCol,
         'x14ac:dyDescent': model.dyDescent,
+        customHeight: model.defaultRowHeight === 0 ? '0' : '1',
```
guyonroche: Could you fix the unit tests that have failed due to this change

autukill: Yes

PR #873, opened 2019-06-24
Uprev fast-csv to latest version which does not use unsafe eval 

miketownsend:
fast-csv@2.4.1 uses a dependency https://www.npmjs.com/package/extended which uses an unsafe eval. 

If we update fast-csv@3.0.1 to the latest version, which has lodash as it's only dependency it should solve this issue: https://github.com/exceljs/exceljs/issues/713


miketownsend:
Closing while I have a look at the test failures.

miketownsend:
The new version of exceljs is breaking the uglify.js minification:

```Unexpected token: keyword «const»````

I'll have another look tomorrow. Any suggestions welcome.

**Comment by miketownsend** [comment]
@guyonroche - any ideas about this?

```
>> Uglifying source ./dist/exceljs.js failed.
Warning: Uglification failed.
Unexpected token: keyword «const».
Line 43131 in ./dist/exceljs.js
 Use --force to continue.
```

The fast-csv module and dependencies are still using "const" which is causing the uglification to fail.

I'm not sure why that one module would not be handled by babel in the same way as the other files. Am I missing something obvious here?

**Comment by paoesco** [comment]
Answer is here : https://davidburgos.blog/how-to-fix-grunt-contrib-uglify-for-es6/

grunt uglify master does not support (yet?) es6 so we need to use the branch harmony.

**Comment by paoesco** [comment]
```
  1 failing
  1) Workbook
       Serialise
         csv file:
     TypeError: csv.createWriteStream is not a function
      at /home/travis/build/exceljs/exceljs/lib/csv/csv.js:96:29
      at new Promise (<anonymous>)
      at module.exports.write (lib/csv/csv.js:88:12)
      at module.exports.writeFile (lib/csv/csv.js:156:17)
      at Context.<anonymous> (spec/integration/workbook/workbook.spec.js:451:10)
      at processImmediate (internal/timers.js:439:21)
```

I guess it's because fast csv has been updated and so certainly we need to change some calls

PR #889, opened 2019-07-08
Add Compression level option to WorkbookWriterOptions for streaming #888

### LINKED ISSUE #888: Add Compression level option to WorkbookWriterOptions for streaming
It is related to https://github.com/exceljs/exceljs/pull/862, we are having some performance issues and we need the same option for `stream.xlsx.WorkbookWriter`

ABenassi87:
Changes:
  - Added zip options for `WorkbookWriterOptions` based on `archiver` docs.
  - Added `testWbStreamCompressionOptions.js` test.
  - Added 'with zip compression option' integration test.
  - Updated `index.d.ts` with the new interfaces.

Note:
https://github.com/archiverjs/node-archiver allows setting the compression level of zip files. I added this option to `stream.xlsx.WorkbookWriter`

Resolves #888

### DIFF
--- a/test/testWbStreamCompressionOptions.js
+++ b/test/testWbStreamCompressionOptions.js
@@ -1,0 +1,102 @@
+const Excel = require('../excel');
+const utils = require('../spec/utils/index');
+
+const filename = process.argv[2];
+console.log(filename);
+const optionsBestCompression = {
+  filename,
+  useStyles: true,
+  zip: {
+    zlib: { level: 9 },// Sets the compression level.
+  },
+};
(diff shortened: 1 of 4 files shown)


test/testWbStreamCompressionOptions.js
```diff
@@ -0,0 +1,102 @@
+const Excel = require('../excel');
+const utils = require('../spec/utils/index');
+
+const filename = process.argv[2];
+console.log(filename);
```
dreamdevil00:
```suggestion
console.log(filename);
```
could be removed

PR #892, opened 2019-07-10
Fix anchor.js

wwojtkowski:
It doesn't work when you have the address with nativeCol = 0

### DIFF
--- a/lib/doc/anchor.js
+++ b/lib/doc/anchor.js
@@ -15,12 +15,12 @@ class Anchor {
       this.nativeColOff = 0;
       this.nativeRow = decoded.row + offset;
       this.nativeRowOff = 0;
-    } else if (address.nativeCol) {
+    } else if (address.nativeCol !== undefined) {
       this.nativeCol = address.nativeCol || 0;
       this.nativeColOff = address.nativeColOff || 0;
       this.nativeRow = address.nativeRow || 0;
       this.nativeRowOff = address.nativeRowOff || 0;
-    } else if (address.col) {
+    } else if (address.col !== undefined) {
       this.col = address.col + offset;
       this.row = address.row + offset;
     } else {


Siemienik:
Col and row are always >= 1

But rigth, that's good pull request

Siemienik:
@wwojtkowski could i please you for write a test that's failed this code before changes?

wwojtkowski:
> Col and row are always >= 1
> 
> But rigth, that's good pull request

I cannot agree with you. Rows are always >= 1 but not col. 
I have Excel's file where I have images in the first column in every row. 
In the Anchor class nativeCol has 0 value but nativeRow has proper value (>=1).
Before my fix, all images from worksheet have nativeRow = 0 and nativeCol = 0 - which is a very serious bug. 
Zrzut ekranu 2019-07-13 o 18 27 15

Siemienik:
@wwojtkowski ok, thanks for the explanation :)

I want to merge it, but it's required to cover changes by tests

vidyakin:
@Siemienik , I have error with the word '**anchor**' when call openFile, may it be the same problem? :)
```
TypeError: Cannot read property 'anchors' of undefined
    at DrawingXform.reconcile (C:\web\my_prj\node_modules\exceljs\lib\xlsx\xform\drawing\drawing-xform.js:94:11)
    at Object.keys.forEach.name (C:\web\my_prj\node_modules\exceljs\lib\xlsx\xlsx.js:103:22)
    at XLSX.reconcile (C:\web\my_prj\node_modules\exceljs\lib\xlsx\xlsx.js:95:33)
    at PromiseLib.Promise.all.then (C:\web\my_prj\node_modules\exceljs\lib\xlsx\xlsx.js:389:16)
```
But it appears in the ONLY Excel file that is important for my task! 
And there are some interesting things I found:
- If I copy all worksheet and paste to new workbook, it works. 
- If I run in Excel "check compatibility" function and do correct operation - it works!

PR #896, opened 2019-07-23
add xml:space="preserve" for all whitespaces

sebikeller:
Currently only spaces at begin or end of a text add the xml:space attribute. 
So at the moment comments with newlines can not be added.


Siemienik:
@sebikeller could I ask you to write a test that covers this change?

guyonroche:
@sebikeller - I've just tried adding some integration tests to master that add white spaces to values and notes at beginning and endings of multiple lines and I can't reproduce your problem.
If you could add an integration test that fails without your change we can review this PR better.
Thanks,

sebikeller:
I agree, with an integration test, its not possible to reproduce the problem.

But if you open the generated file with excel, you'll see the difference.

## branch master
master
[exceljs-master.xlsx](https://github.com/exceljs/exceljs/files/3736814/exceljs-master.xlsx)

## applied pr
out pr896
[exceljs-pr896.xlsx](https://github.com/exceljs/exceljs/files/3736816/exceljs-pr896.xlsx)

sebikeller:
Recreating the file in excel, following xml can be extracted:

``` xml
<comments xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
  <authors>...</authors>
  <commentList>
    <comment ref="A1" authorId="0" shapeId="0">
      <text>
        <r>
          <rPr>
            <b/>
          </rPr>
          <t>First Line:</t>
        </r>
        <r>
          <t xml:space="preserve">
</t>
        </r>
        <r>
          <t xml:space="preserve">Second Line
</t>
        </r>
        <r>
          <t xml:space="preserve">Third Line</t>
</t>
        </r>
        <r>
          <t>Last Line</t>
        </r>
      </text>
    </comment>
  </commentList>
</comments>

```

guyonroche:
@sebikeller  - agreed! I'll add a unit test on the strings xforms to keep a check on this...

PR #906, opened 2019-07-31
Exclude Infinity on createInputStream

sophiedophie:
Character such as `6E10000` becomes Infinity with `Number()`
but it's not `NaN`, so just returns it.
It causes error for excel file so can't open generated file.

We need to check if it's Infinity or not.


Siemienik:
@sophiedophie thanks for the pull request :), could I ask you to write a test that covers this change?

sophiedophie:
@Siemienik Added the test. Please take a look for it and let me know if something need to be improved.

PR #907, opened 2019-07-31
Feature/Sheet Protection

karabaesh:
Changes:
- New feature: Ability to protect and unprotect worksheets


lib/doc/worksheet.js
```diff
@@ -109,8 +110,12 @@ class Worksheet {
     // for images, etc
     this._media = [];
 
+    // worksheet protection
+    this.sheetProtection = {};
```
guyonroche: Since sheet protection is an all-or-nothing feature - it would be better to initialise this to null. This would make the rendering and parsing easier.

lib/doc/worksheet.js
```diff
@@ -579,6 +584,29 @@ class Worksheet {
     return image && image.imageId;
   }
 
+  // =========================================================================
+  // Worksheet Protection
+  protect(password, options){
```
guyonroche: I know - thanks to Microsoft's "special" implementation of password hashing that it's hard to implement this function in an async way - but I think it would be worth wrapping this function in a Promise.
Calculating the hash takes almost a whole second and this would be a good place to use worker threads if available.

lib/doc/worksheet.js
```diff
@@ -579,6 +584,29 @@ class Worksheet {
...
+        this.sheetProtection.saltValue = Encryptor.randomBytes(16).toString('base64');
+        this.sheetProtection.spinCount = 100000;
+        this.sheetProtection.hashValue = Encryptor.convertPasswordToHash(password, 'SHA512', this.sheetProtection.saltValue, this.sheetProtection.spinCount);
+    }
+    if(options){
+      this.sheetProtection = Object.assign(this.sheetProtection, options);
+    }
+  }
+
+  unprotect(){
```
guyonroche: This can just assign sheetProtection to null

lib/utils/encryptor.js
```diff
@@ -0,0 +1,62 @@
+'use strict';
+
+const crypto = require('crypto');
+const buffer1 = require('buffer');
```
guyonroche: "Buffer" is native JS - no need to require it

lib/utils/encryptor.js
```diff
@@ -0,0 +1,62 @@
...
+   * @param {string} algorithm - The hash algorithm.
+   * @param {Array.<Buffer>} buffers - The buffers to concat and hash
+   * @returns {Buffer} The hash
+   */
+  hash(algorithm) {
+    const buffers = [];
+    for (let i = 1; i < arguments.length; i++) {
+      buffers[i - 1] = arguments[i];
+    }
+    algorithm = algorithm.toLowerCase();
```
guyonroche: There's no need to execute this line (and the following 4 lines) every time we call hash(). This check should really move down to the convertPasswordToHash() function

lib/utils/encryptor.js
```diff
@@ -0,0 +1,62 @@
...
+const buffer1 = require('buffer');
+
+const Encryptor = {
+  /**
+   * Calculate a hash of the concatenated buffers with the given algorithm.
+   * @param {string} algorithm - The hash algorithm.
+   * @param {Array.<Buffer>} buffers - The buffers to concat and hash
+   * @returns {Buffer} The hash
+   */
+  hash(algorithm) {
```
guyonroche: If the function signature were hash(algorithm, ...buffers), we can eliminate the next 4 lines

lib/doc/worksheet.js
karabaesh: Wrapped this into a promise and changed commit() in worksheet-writer. I cannot determine any improvements.

lib/utils/encryptor.js
karabaesh: Can you make a suggestion. Could not get your point.

lib/doc/worksheet.js
guyonroche: Cool - it's not really about improvements but future proofing the API. In the case that one day we might be able to make this call properly asynchronous. 700 milliseconds is a very long time to be stuck adding a password hash to a sheet and if we can pass that over to a worker thread it will be much better for the app as a whole.

lib/utils/encryptor.js
guyonroche: The point is - this hash() function is called 100000 times by convertPasswordToHash(). The expression hashes.indexOf(algorithm) is running a linear array search every single time and is unnecessary if the check is performed within the convertPasswordToHash() function.

karabaesh: Ok I see your point. Change has been implemented.

PR #923, opened 2019-08-17
Add document for zip options of streaming WorkbookWriter

piglovesyou:
@ABenassi87 Added doc for you change.

Ref. #889

### DIFF
--- a/README.md
+++ b/README.md
@@ -1833,6 +1833,7 @@ The constructor takes an optional options object with the following fields:
 | filename         | If stream not specified, this field specifies the path to a file to write the XLSX workbook to. |
 | useSharedStrings | Specifies whether to use shared strings in the workbook. Default is false |
 | useStyles        | Specifies whether to add style information to the workbook. Styles can add some performance overhead. Default is false |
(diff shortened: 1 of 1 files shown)


alubbe:
@guyonroche how do we deal with translating this to Chinese? Do we keep this PR open until someone adds a Chine translation or do we merge the PR and create a ticket for its Chinese translation?

ABenassi87:
@alubbe I could translate to Spanish too if you want that

piglovesyou:
Why don't we discuss how exceljs makes docs and how we translates it?

* If every time we discuss docs and translation at the same time, documenting cost will go high
* If documenting cost goes high, everyone stops documenting

I think documenting in English first is a good idea in order to supply a minimum and essential information fast. At the moment we can put `// TODO` in corresponding position in README_zh.md not to forget the later work.

guyonroche:
@alubbe, @ABenassi87, @piglovesyou - translations (for Chinese so far) have come from volunteer submissions.
I (and quite a few contributors) try to keep the main English README up to date so it does tend to change with some frequency. This does mean the translations risk getting stale. I did mention this to the original Chinese contributor and he said he was happy to keep it up to date.
I expect the best option is to add a little caveat above the translations explaining this.

@ABenassi87 - by all means, if you want to add a Spanish version then it will be most welcome.

PR #927, opened 2019-08-20
Fix typing when Typescript is in strict mode

lbguilherme:
ExcelJS 1.15.0 causes a type error on any project that depend on it and is in strict mode. Downgrading to 1.14.0 solves the issue.

```
node_modules/exceljs/index.d.ts(1154,2): error TS7010: 'protect', which lacks return-type annotation, implicitly has an 'any' return type.
node_modules/exceljs/index.d.ts(1155,2): error TS7010: 'unprotect', which lacks return-type annotation, implicitly has an 'any' return type.
```

This PR fixes the issue.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1151,8 +1151,8 @@ export interface Worksheet {
 	/**
 	 * Worksheet protection
 	 */
-	protect(password: string, options: Partial<WorksheetProtection>);
-	unprotect();
+	protect(password: string, options: Partial<WorksheetProtection>): void;
+	unprotect(): void;
 }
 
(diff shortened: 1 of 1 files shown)


eugene1g:
Good catch! 
There is another PR for this at https://github.com/exceljs/exceljs/pull/926 (note that `protect` returns a Promise as per https://github.com/exceljs/exceljs/blob/d29d2ecf8399cb2001e7410ac5b2dc330bf6f30f/lib/doc/worksheet.js#L593)

lbguilherme:
Oh, awesome!

I'm closing this in favor of #926

PR #929, opened 2019-08-21
 Modify code causes richtext and comments to be lost

zhangyu836:
Modify code causes richtext and comments to be lost in resulted file.
Due to the way exceljs handles comments, comments in empty cells will always be lost.


Siemienik:
what is the difference to https://github.com/exceljs/exceljs/pull/976 ?

PR #930, opened 2019-08-21
Update index.d.ts

cosmonovallc:
error TS7010: 'protect', which lacks return-type annotation, implicitly has an 'any' return type


Siemienik:
it is not a `void`:
https://github.com/exceljs/exceljs/blob/f50693b344c938d4384208a9ead8bdfa9da17600/lib/doc/worksheet.js#L593

alubbe:
@cosmonovallc could you amend this PR to include the right return type?

PR #933, opened 2019-08-23
array formula

### LINKED ISSUE #932: broken array formula
when the loaded file contains array formula, they are viewed as classic formula and therefore the saved file is broken

to reproduce, simply create a file with an array formula with excel, load it using exceljs, save it and open the produced file using excel: the array formula is viewed as a classic one and therefore is invalid

yoann-antoviaque:
fix #932 :)


alubbe:
thanks for opening this pr! Could you please add a test case?

guyonroche:
This is a good change, but there are a few issues - as ell as tests, there also needs to be a means in the API to add/edit shared/array formulas. I'm looking into this now...

PR #936, opened 2019-08-27
Add "sum" subtotal function

egmen:


alubbe:
@guyonroche we have no tests on this right? So this PR could be merged as-is?

PR #968, opened 2019-09-17
Added comment-getter to the Cell class

yungcatx:
As i didn't get any response in 2 weeks, i've done it by myself. I don't know if this approach is suitable for your architecture, but i didn't understand why this._comment property is undefined in Cell class. Also checked all of X-form things related to comments, everything is fine there.

### DIFF
--- a/lib/doc/cell.js
+++ b/lib/doc/cell.js
@@ -224,6 +224,10 @@ class Cell {
     return this._comment && this._comment.note;
   }
 
+  get comment() {
+    return this._value.model.comment;
+  }
+
   set note(note) {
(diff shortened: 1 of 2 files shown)


guyonroche:
@yungcatx apologies for the delay - I tend to prioritise pull requests over issues as they indicate a higher level of need.
Regarding this PR - access to the cell comment is already available via that 'note' property which returns the model of the attached comment. The reason we called this 'note' is that Excel has changed the meaning of 'comment' to a chat-style comment where users can hold a discussion and the old-style comment has been renamed to 'note'. So far, ExcelJS only supports the old-style notes

Issue #951, opened 2019-09-06
Force formula re-calculation on file open from Excel


zymon:
Hello,
I desperately need support for forcing excel to re-calculate the formulas on open.

This is following part:
<x:calcPr calcId="999999" fullCalcOnLoad="1"/>

Can you give me some clues where to add it, so I could contribute and add this functionality.

guyonroche:
Added in v3.3.0

PR #980, opened 2019-10-03
#951: Force formula re-calculation on file open from Excel

zymon:
Added Calculcation Properties to Workbook and implemented generation of  'fullCalcOnLoad' attribute in order to force Excel to recalculate all formulas when workbook is open.


cederigo:
Whats the state here? Any work left?
Im also very interested in this feature.

guyonroche:
@zymon Thanks for submitting this great PR - docs, tests, everything! Only comment is please don't modify the package version - this is managed during releases and is also likely to hit merge collisions.
Travis is green, will merge

PR #1000, opened 2019-10-26
Add headerFooter to worksheet model when importing from file

### LINKED ISSUE #999: Can't import headerFooter from existing xlsx files
Hello all,

Since #863 (v1.13.0), header and footer can be exported, which I think is a great feature!
However, I found that header and footer cannot be imported from existing xlsx files.

To reproduce, create an xlsx file with any header/footer (say foo.xlsx) and execute in NodeJS:
```js
const Excel = require('exceljs')
const wb = new Excel.Workbook()
wb.xlsx.readFile('foo.xlsx')
wb.getWorksheet(1).headerFooter
```
This returns just `undefined`, but I expect an object with properties like `oddFooter`.

I opened a PR for this: #1000

kigh-ota:
Fix #999 

I just added `headerFooter` to the model of `WorkSheetXform` in `#parseClose()`, in order that `Worksheet` object has `headerFooter` property read from existing xlsx file.

Even when `sheetX.xml` doesn't contain `<headerFooter>` node, `Worksheet` object still has `headerFooter` property with value `null`; I wonder whether this behavior is good.

Please let me know if anything is needed for this PR.

### DIFF
--- a/lib/xlsx/xform/sheet/worksheet-xform.js
+++ b/lib/xlsx/xform/sheet/worksheet-xform.js
@@ -313,6 +313,7 @@ class WorkSheetXform extends BaseXform {
           properties,
           views: this.map.sheetViews.model,
           pageSetup,
+          headerFooter: this.map.headerFooter.model,
           background: this.map.picture.model,
           drawing: this.map.drawing.model,
           tables: this.map.tableParts.model,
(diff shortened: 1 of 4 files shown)


guyonroche:
@kigh-ota cool - nice catch. Will merge...

kigh-ota:
@guyonroche 
Many thanks for the quick response!

PR #1007, opened 2019-10-29
Rename .browserslintrc.txt to .browserslistrc

takenspc:
`.browserslitrc.txt` is not recognized by browserslist. The correct file name is `.browserslistrc`.

https://github.com/browserslist/browserslist#queries
> `.browserslistrc` config file in current or parent directories.


takenspc:
Thank you for reviewing and merging.

PR #1016, opened 2019-11-03
Feature/stream writer add background images

brunoargolo:
Implemented background images for sheets on the stream writer, obeying the same API as the the regular workbook/worksheet. This does not include regular images (drawings), only background (picture).

please note: there was a small block of code before my PR that tried to deal with background images but it never really worked, it was probably some left over. I have removed it to avoid confusion.


goums:
hi @brunoargolo @guyonroche 
thanks a lot for this feature!
I'm looking to implement addImage feature also on the stream workbook api.
have you guys started working on this already?
any clues on how I should get started
Thanks in advance for your help

PR #1018, opened 2019-11-04
Large excels - optimize performance of writing file by excelJS + optimize generated file (MS excel opens it much faster)

pzawadzki82:


lib/xlsx/xform/style/styles-xform.js
```diff
@@ -82,6 +82,8 @@ class StylesXform extends BaseXform {
     // add default fills
     this._addFill({type: 'pattern', pattern: 'none'});
     this._addFill({type: 'pattern', pattern: 'gray125'});
+
+    this.weakMap = new Map()
```
guyonroche: Nice catch! Quick question - any good reason not to use a WeakMap here?

lib/xlsx/xform/sheet/data-validations-xform.js
```diff
@@ -28,17 +28,65 @@ function assignBool(definedName, attributes, name, defaultValue) {
   }
 }
 
+function mergeDataValidations(model) {
```
guyonroche: Could you add a comment describing how the model is being optimised here?

lib/xlsx/xform/style/styles-xform.js
pzawadzki82: No reason. i'll change it

lib/xlsx/xform/sheet/data-validations-xform.js
pzawadzki82: currently excelJS model stores dataValidation per each cell. So if there is many cells data validation is created for every cell. I'll give you example: on my project there 5k rows x 20 columns ~ 100k cells. So in this case MS Excel openes generated file very long (minutes) or if some number is exceeded it reports that ms excel is corrupted.
In many cases validation is the same for bigger range (in our case it is while column but header). So my optimization just replaces multiple dataValidation xml nodes(with the same formula and continous range) with with one where address is defined as range instead one cell. So on my project instead 100k dataValidation nodes there is 20 and this file is opened by MS Excel fast and without errors.

lib/xlsx/xform/style/styles-xform.js
pzawadzki82: Changed

guyonroche:
@pzawadzki82 thanks for this PR - I had to modify it a little so that it would handle parsing the files too and I've extended it to include rectangular ranges

Issue #991, opened 2019-10-24
read csv file issue


hehhuang:
var Excel = require("exceljs");
// read from a file
var workbook = new Excel.Workbook();
workbook.csv.readFile("3.csv")
  .then(worksheet => {
    // use workbook or worksheet  
    console.log(worksheet.getColumn(1).values);
  });
/*output :
[ <1 empty item>,
  'name',
  2001-04-30T16:00:00.000Z,
  2001-04-30T16:00:00.000Z,
  'george\'s Grocery aa55' ]*/
**but the csv file is like this:**
name
George's Grocery 005
Quinn's Supermarket #005
george's Grocery aa55
the issue is that if the value contains space+number(or other Special symbol), it will output a date type value.For example "a 123" will output "0122-12-31T15:54:17.000Z". How to resolve this.(xlsx file can work well).

LibertyNJ:
I am encountering a similar issue, except it does not appear to be tied to a space or a special symbol. 

String 0020Z6TNY parses as a string: 0020Z6TNY. (expected).
String 0020Z6TRL parses as a string: 0020Z6TRL. (expected).
String 00210PRG1 parses as a date: 1920-12-01T06:00:00.000Z. (unexpected).
String 00210PRGS parses as a date: 1920-12-01T05:00:00.000Z. (unexpected).
String 00211MDN2 parses as a date: 1921-01-01T07:00:00.000Z. (unexpected).

I wonder if this could be related to day.js date validation?

As a workaround I passed in an [options object that specifies how each column should be parsed](https://github.com/exceljs/exceljs#reading-csv). This allowed me to override the behavior on a column by column basis. Changing the dateFormats part of the options object does not appear to work, which again may point back at day.js.

LibertyNJ:
I looked into it a little bit more, and this is what I found. I would submit a pull request but there are a few different things going on between exceljs and dayjs, and I'm not sure what would be the right approach to rectifying it.

CSV.createInputStream's default map attempts to convert a cell's value into a date with dayjs. If the result passes a test as a valid date, it is returned as a date object for the value for the cell. However, the implementation of dayjs in exceljs is not extended with the customParseFormat plugin. Another issue is that it looks like dayjs does not accept an array of formats as an argument, only a single string as a primitive or a property in an object.

[...]

Issue #794, opened 2019-04-24
Make package.json friendly


cncolder:
1. Move `@types/node` into devDependencies.
2. Do not pin jszip version `"jszip": "^3.1.3"`.
3. Consider replace moment with [dayjs](/iamkun/dayjs). It's too heavy since we only parse time string.
4. add `browser` field.

Siemienik:
@cncolder  Could you open PR with these changes?

A1rPun:
1. Is done already
2. Pinning on major version is good practice, why removing this caret for `jszip` specifically?
3. This is a must. I wanted to make use of this library until I saw `moment` in the `package.json` which is [33% of your current codebase](https://bundlephobia.com/result?p=exceljs@1.14.0)!
4. Is done already

havietduc91:
Hi exceljs team, 

Thank a lot for effort you've put in this library 
I'd like to replace `moment` with `dayjs` to make bundle size more lightweight.
So could you please replace it.

Many thanks.

PR #1019, opened 2019-11-04
Fix issue # 991

LibertyNJ:
There is an issue where the csv.read/readFile method will incorrectly parse certain strings as dates that should not be dates. This may stem from differences in the API of momentjs and dayjs that were not addressed after migration. 

By default dayjs will transform any string that matches the regex /^\d{4}-?\d/ into a valid date, whether it makes sense for that string to be interpreted as a date or not. This appears to be intended behavior on the part of dayjs, as it lets the module parse dates from strings in non-standard formats. If this is not the intended behavior of the csv.read/readFile methods, the default options.map method needs some more logic to help decide when to return a date.


damianaiamad:
I'm a bit late to this party, sorry. I think I've identified a further problem with dayjs that means this fix does not fully solve the problem.

Please see https://github.com/exceljs/exceljs/issues/991

LibertyNJ:
Can confirm, 'D/M/YY' pattern breaks test with 'NZ1019316'. Dayjs appears to have further issues as @damianaiamad identified in #991.

Reverting to moment is definitely a solution (wasn't broke, why fix it?), if the size of that module is not still a concern as it was in issue #794.

damianaiamad:
One advantage of @LibertyNJ's  solution is that it allows turning off automatic date parsing as follows:

`await workbook.csv.read(stream, {dateFormats: []})
`

Issue #1008, opened 2019-10-30
"Zone.js has detected that ZoneAwarePromise `(window|global).Promise` has been overwritten


taruntadikonda:
There is error when exceljs is used in code.

Error:

**Error: "Zone.js has detected that ZoneAwarePromise `(window|global).Promise` has been overwritten.
Most likely cause is that a Promise polyfill has been loaded after Zone.js (Polyfilling Promise api is not necessary when zone.js is loaded. If you must load one, do so before loading zone.js.)"
    Angular 5
    o exceljs.min.js:3**

My service.ts file

`import { Injectable } from '@angular/core';`
`import * as Excel from "exceljs/dist/exceljs.min.js";`
`import * as fs from 'file-saver';`
`@Injectable({`
  `providedIn: 'root'`
`})`
`export class ExcelserviceService {`

 ` constructor() { }`

  `generatereport(dummydata):void{`
    
    let workbook = new Excel.Workbook();
   ` let worksheet = workbook.addWorksheet('Car Data');`
    
 ` }`
`}`

nvkolluru:
I am able to make it work in local by removing zone.js imports from polyfills.ts file and adding them to main.ts at the end. But, having same ZoneAwarePromise issue again after build and publish. Any updates ?

bleuscyther:
No solutions worked for me i saw some old libraries with this issue,  i could not see exactly what they changed to fix it ( 2017-18)

nvkolluru:
Made it work for the normal build (ng build), but for PROD Build - no luck yet.

inyourdrimz:
fixed mine by importing it before 'core-js' and 'zone.js' in the pollyfills.ts file

pollyfills.ts file:

```
import 'exceljs/dist/exceljs.min.js';

/** IE9, IE10 and IE11 requires all of the following polyfills. */
import 'core-js/es6/symbol';
.
.
. 
.
.
.
import 'zone.js/dist/zone';
```

bleuscyther:
> fixed mine by importing it before 'core-js' and 'zone.js' in the pollyfills.ts file
> 
> pollyfills.ts file:
> 
> ```
> import 'exceljs/dist/exceljs.min.js';
> 
> /** IE9, IE10 and IE11 requires all of the following polyfills. */
> import 'core-js/es6/symbol';
> .
> .
> . 
> .
> .
> .
> import 'zone.js/dist/zone';
> ```

[...]

PR #1030, opened 2019-11-20
     remove core-js/ import

bleuscyther:
This affects angular. Because core-js is imported inside the library , the Promise  polyfill is in conflict with  zone.js. Since Excel.js puts browsers compatibility on the hands of the devs. ([link](https://github.com/exceljs/exceljs#es5-imports)) , we can add it manually in a way that does not conflict with Angular.

this PR should fix : https://github.com/exceljs/exceljs/issues/1008

This was fixe din many other package and is referenced here:
 angular/zone.js#783

### DIFF
--- a/lib/exceljs.browser.js
+++ b/lib/exceljs.browser.js
@@ -1,7 +1,4 @@
 /* eslint-disable import/no-extraneous-dependencies,node/no-unpublished-require */
-require('core-js/modules/es.promise');
-require('core-js/modules/es.object.assign');
-require('core-js/modules/es.object.keys');
 require('regenerator-runtime/runtime');
 
 const ExcelJS = {


guyonroche:
@bleuscyther - I'm tending to agree - though I might go one step further and remove the regenerator-runtime/runtime require as well, leaving the whole task of polyfill support to the app in question. What do you think?

bleuscyther:
that would be perfect.

guyonroche:
@bleuscyther I've just merged - I'm going to make one further change... In order to avoid breaking changes, I'm going to copy exceljs.browser.js to exceljs.bare.js that removes all polyfills and restore exceljs.browser.js to keep them. To include the exceljs bundle without pollys, just change the import from dist/exceljs.min.js to dist/exceljs.bare.min.js

bleuscyther:
Nice i am going to test this week

bleuscyther:
There was a new error:
> Uncaught ReferenceError: regeneratorRuntime is not defined

from what i see this is a babel issue related to how the file was built. https://github.com/babel/babel/issues/9849

As a quick fix, I installed the missing package : 

`npm install regenerator-runtime`
  
and put the import in the polyfill for now:
```typescript
import "regenerator-runtime/runtime";
```


Using angular i had to do the usual additional changes (tested on version 8), it works fine after that

``` javascript
// ./tsconfig.json
...
  "module": "esnext",
    "moduleResolution": "node",
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "importHelpers": true,
    "allowSyntheticDefaultImports": true,
    "target": "es2015",
   "paths": {
        "exceljs": [
          "../node_modules/exceljs/dist/exceljs.bare"
      ],
..
```
Check this [(link)](https://angular.io/guide/typescript-configuration) for additional configs and updated values/fields.

``` javascript
...
// ./angular.json
...
"architect": {
        "build": {
          "builder": "@angular-devkit/build-angular:browser",
          "options": {
                ...
               "scripts": [
                  ...
                  "./node_modules/exceljs/dist/exceljs.bare.js"
               ]
               ...
 ```
``` typescript
//my .component.ts

import * as ExcelJS from 'exceljs/dist/exceljs.bare.js';
import {Row, Workbook, Worksheet} from 'exceljs';
...
    const workbook: Workbook = new ExcelJS.Workbook(); // Important don' t use Class directly the wrong module will be loaded
```

guyonroche:
@bleuscyther 
Great you got it working - does this leave any changes required for ExcelJS?
My thoughts are that the 'bare' option leaves all the polyfill dependencies for the app to fetch - including "regenerator-runtime/runtime" - though I should probably add this to the readme

bleuscyther:
Probably add it on the Readme. Thank you for your help.

PR #1038, opened 2019-11-27
writeBuffer work again in browser when in production

### LINKED ISSUE #1014: Problem exporting Excel with Angular in production
Exprting data in Excel in developement environment work very well but in production silently error occur. This problem seem to be recurent from many version. Actually we are with ExcelJS version 3.1.0 We try ExcelJS 3.3.1 and we have the same problem. We can't rollback version to old version because we use fonctionnality contained in only new version. We expect correction soon. The problem seem to be with  **workbook.xlsx.writeBuffer()**.then

Thanks

manuelblum:
Fix browser path in package.json so workbook.xlsx.writeBuffer() work again in browser.

Should fix #1014 

**Problem:**
The [current package.json](https://github.com/exceljs/exceljs/blob/master/package.json#L19) points to `"browser": "./dist/exceljs.min.js",` but there is no such file in the distributed dist directory from version 3.4.0. 

Screenshot 2019-11-27 at 10 13 05

**Solution:**
I think this path should point to: `"browser": "./dist/exceljs.js",`

**other possible Solutions**
Build `exceljs.min.js` again

Why are you not building "./dist/exceljs.min.js" anymore?

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "node": ">=8.3.0"
   },
   "main": "./lib/exceljs.nodejs.js",
-  "browser": "./dist/exceljs.min.js",
+  "browser": "./dist/exceljs.js",
   "types": "./index.d.ts",
   "files": [
     "dist",


guyonroche:
Hi @manuelblum 
I'm not sure how you encountered a published version without a dist/exceljs.min.js file. It's always been built as far as I know - and I just checked v3.5.0 and it's still there. Could you recheck?

Siemienik:
@guyonroche in issues this bug occurs many times. Something works wrong there.

guyonroche:
I'll need an example that reproduced the issue. I just npm installed exceljs@3.5.0 and ran the following JS fine:
```javascript
const browserExcel = require('exceljs/dist/exceljs.min.js');
console.log('browserExcel', browserExcel)
```

guyonroche:
@manuelblum - I'm not convinced this PR fixes the linked issue. Also - from the image you have included above, it appears that the exceljs.min.js file does exist, but that your IDE has hidden it. The triangle to the left of the exceljs.js file indicates related files have been hidden - if you click on the triangle, I suspect you will see the exceljs.min.js file appear.
Can you confirm? If so then we should close this PR

manuelblum:
Your are right, there is a min.js file.

PR #1047, opened 2019-12-12
Clarify merging cells by row/column numbers

kendallroth:
The previous examples for merging cells by row/column numbers were confusing and didn't actually describe what parameters the function was expecting. The example made it appear that `mergeCells(10, 11, 12, 13)` was the same as `mergeCells("G10", "H11")`, which is not the case. This example has been updated as well as another comment added.


Siemienik:
for me, both versions are clear to understand. 
If I have to choose one, I'll prefer "top-left-bottom-right"

PR #1058, opened 2019-12-20
fix typings for cell.note

xydens:
There were issue that cell comments are accessible by Cell.note property, but in index.d.ts were "comments" instead of "note".


Siemienik:
Could you remove version increasing? it's not required here, scripts care about that 😸 

ofc, You have right with typings :)
https://github.com/exceljs/exceljs/blob/master/lib/doc/cell.js#L222

Issue #1064, opened 2020-01-02
This workbook.addWorksheet () should handle issues caused by case of worksheet names


Alanscut:
### Wrong reason:
The sheet name of the excel file created in the windows system is not case sensitive. But the ```workbook.addWorksheet ()``` method does not do the corresponding processing.

### For example: 
 the following code can successfully create two worksheets, but the actual generated file has problems.
```javascript
var worksheet = workbook.addWorksheet("sheet1");
var worksheet2 = workbook.addWorksheet("Sheet1");
```

PR #1065, opened 2020-01-02
Fix problems caused by case of worksheet names

Alanscut:
### Changes

1.The sheet name of the excel file created in the windows system is not case sensitive. But the ```workbook.addWorksheet ()``` method does not do the corresponding processing, This PR fixes issues caused by case of sheet names.

2.fix for #1064

### DIFF
--- a/lib/doc/workbook.js
+++ b/lib/doc/workbook.js
@@ -51,7 +51,7 @@ class Workbook {
       console.warn(`Worksheet name ${name} exceeds 31 chars. This will be truncated`);
     }
     name = (name || `sheet${id}`).substring(0, 31);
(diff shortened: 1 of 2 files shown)


Siemienik:
@Alanscut Is will still working when it will attempt to read file with two worksheets named: "name", "NAME" ?

Alanscut:
@Siemienik yes, it could. But when we create a file with two worksheets with the same names regardless of case, the file will report an error when opening it. So when we prevent creating such file, the case you mentioned will not exist.

guyonroche:
@Alanscut - nice choice of 31 character sheet name in the spec file :-)

Issue #1045, opened 2019-12-09
horizontalCentered & verticalCentered in page not working


tinhnqjp:
Hi all, plz help me

**Template file:**
![image](https://user-images.githubusercontent.com/34364630/70420419-0bec8080-1aab-11ea-9e5e-c96268b91f32.png)
Center on page: 

- horizontally : Checked
- vertically : Checked

**Source coder:**

```
worksheet.pageSetup.horizontalCentered = true 
worksheet.pageSetup.verticalCentered = true
```

**File output:**
- horizontal : uncheck
- vertical : uncheck

![image](https://user-images.githubusercontent.com/34364630/70420596-6980cd00-1aab-11ea-8b30-7933fb9016b3.png)

LiaoZhiyong:
worksheet-xform.js
line: 216
const printOptionsModel = {
      showRowColHeaders: model.pageSetup.showRowColHeaders,
      showGridLines: model.pageSetup.showGridLines,
      horizontalCentered: model.pageSetup.horizontalCentered,
      verticalCentered: model.pageSetup.verticalCentered,
    };

line: 238
this.map.printOptions.render(xmlStream, printOptionsModel);
this.map.pageMargins.render(xmlStream, pageMarginsModel);

Siemienik:
I will be glad for pull request with fixation 😸

PR #1073, opened 2020-01-15
fix issue #1045 horizontalCentered & verticalCentered in page not working

Alanscut:
### Changes 
1. This PR solves the problem that horizontalCentered & verticalCentered is not working.

2. Fix for #1045


Siemienik:
@Alanscut thanks for that PR, it's a bit of great work here:) 
Could I ask you to write at least one test which checks this logic? (for example, write and read file + asserts for these properties).

Alanscut:
hi @Siemienik 
Thank you very much for your comment.
See https://github.com/exceljs/exceljs/blob/master/spec/unit/xlsx/xform/sheet/print-options-xform.spec.js#L23

Siemienik:
hmm.. so how the tests work ok before and now after source code was changed? Are both versions correct?

Alanscut:
@Siemienik Having researched the unit test, I found I have misunderstood it. Actually the testcases for it lies in file `worksheet-xform.spec.js`, property `horizontalCentered` are not tested before, so I have add the test case and fixed issue #1079 also.
https://github.com/exceljs/exceljs/blob/9054c3240ea4f5fa70720bfedf14b0c9da751e71/spec/unit/xlsx/xform/sheet/worksheet-xform.spec.js#L53-L70

Siemienik:
great :)

Issue #839, opened 2019-05-29
Duplicate or copy row together with stylings and data


oalexandru:
Hello,

Can anybody tell me how can I duplicate a row in a worksheet(I aim for identical rows, meaning stylings, borders, merged cells, data etc).I couldn't find a way.

P.S. What I actually want to achieve is duplicating a row in the middle of the spreadsheet, so that all the rows below are shifted by one row towards the bottom, but I could use only the duplication part, and I can carry on with the shifting afterwards.

Thanks:)

dogusev:
@oalexandru 

https://github.com/exceljs/exceljs#rows

I think this is something you are looking for
```
// remove one row and insert two more.
// Note: rows 4 and below will be shifted down by 1 row.
var newRow3Values = [1,2,3,4,5];
var newRow4Values = ['one', 'two', 'three', 'four', 'five'];
worksheet.spliceRows(3, 1, newRow3Values, newRow4Values);
```

Gricardov:
I'm having the same problem. I want to shift down the content without losing styles :(

oalexandru:
Managed to copy the style, height, borders and values from one row to the other pretty easily with:

currentCell.style = cellToBeShifted.style;
currentCell.value = cellToBeShifted.value;
currentCell.height = cellToBeShifted.height;

but didn't find a way of copying the merging model, so I ended up removing all the merges from the document and applied merging manually after I did the shifting with

worksheet.mergeCells()

PR #1078, opened 2020-01-15
function duplicateRows added

cbeltrangomez84:
I added a function to duplicate rows with style and values in a worksheet. It still needs to be improve since formulas are not copied and it does not support merged cells, this function was created to support issue #839


Siemienik:
@cbeltrangomez84  something went wrong during testing

cbeltrangomez84:
Sorry @Siemienik I don´t know how to do it. 😳
I sent you an email.

Siemienik:
@cbeltrangomez84 ok, each important thinks has been done:),

If I ask You, please write some documentation to README how to use this function :)

lib/doc/worksheet.js
```diff
@@ -344,6 +344,49 @@ class Worksheet {
...
+  duplicateRow(start, count) {
+    // I want to add after the start row
+    start++;
+
+    const nKeep = start;
+    const nEnd = this._rows.length;
+    let i;
+    let rSrc;
+
+    // insert new cells
```
guyonroche: Thinking out loud here - this for loop is a duplication of code already implemented in worksheet.spliceRows()
Perhaps a better strategy would be to prepare an array of insertion values and let spliceRows() do the donkey work...
```javascript
const values = this._rows[start - 1].values;
const inserts = new Array(count - 1).fill(values);
const deleteCount = insert ? 0 : count - 1;
this.spliceRows(start + 1, deleteCount, ...inserts);

// copy styles from start row to cloned rows
...
```

cbeltrangomez84:
Thanks @guyonroche and @Siemienik 

I made another pull request with the changes and README (plus another test)

PR #1087, opened 2020-01-21
Prevent from unhandled promise rejection durning workbook load

sohai:
Since https://github.com/exceljs/exceljs/pull/829 `StreamBuf.write` returns Promise, that leads to unhandled promise rejection in `Workbook.load` because `ZipStream.write`do not expect `StreamBuf.write` to return Promise.


alubbe:
Good catch!

PR #1088, opened 2020-01-21
Duplicate rows

cbeltrangomez84:
Thanks @Siemienik and @guyonroche for the suggestions.

I added the explanation in README as asked.

Regards


guyonroche:
@cbeltrangomez84 oops - my apologies, after I saw the original PR was merged, I went ahead and refactored it. I'll resolve the collisions so we can include your README changes. Thanks

cbeltrangomez84:
Hi @guyonroche thanks for refactoring the code, one test is failing because the default value for the 'insert' argument in my code was false. Since the refactor is true it doesn`t match the test. 

Could you change the argument? or if you prefer I'd change the test.

Thanks

PR #1089, opened 2020-01-22
Update TS definition to reference proper internal libraries

jakawell:
### Summary
* Replace empty "Stream" interface with the Node internal "stream.Stream" interface
* Replace empty "Writable" interface with the Node internal "events.EventEmitter" interface

### Notes
When using the library with TypeScript, I noticed that a couple of empty interfaces were causing errors in the TS parsing. I checked the source code and verified the correct internal interfaces so that the typings are accurate.

I know unit tests are required for PRs, but I don't see anywhere in this project that it has unit tests for the TypeScript definition, and I'm not really sure how to do that (or if it would even be useful). Let me know if I'm wrong!

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,4 @@
 declare interface Buffer extends ArrayBuffer { }
-declare interface Stream { }
-declare interface Writable { }
 
 export const enum RelationshipType {
 	None = 0,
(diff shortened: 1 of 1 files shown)


guyonroche:
@jakawell thanks for the fix.
Regarding tests - I agree it would be useful to have a test that validates the type files. I think it could work by adding typescript, et al, as dev dependencies and having a new flavour "test:typescript" that would only work if the types parsed correctly. I'll see if I can make some that fail before merging this PR

guyonroche:
@jakawell - I've made a little progress. I've added a new test folder: spec/typescript with a couple of tests that execute ok but now they fail due to typescript compiler errors.
See: spec/typescript/exceljs.spec.ts - If I change the code to use:
const stream = wb2.xlsx.createInputStream() as any;
the test runs ok, but take away the "as any" and it fails. Do you know what is wrong?

PR #1092, opened 2020-01-26
Fix index.d.ts

Siemienik:
https://github.com/exceljs/exceljs/blob/9ec41e360d1700cd482e5391bac791db84f42c74/lib/doc/cell.js#L671

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -388,7 +388,7 @@ export interface CellModel {
 	text?: string;
 	hyperlink?: string;
 	value?: CellValue;
-	master: Cell|string;
+	master: string;
 	formula?: string;
 	sharedFormula?: string;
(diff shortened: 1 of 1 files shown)


Siemienik:
I have the second bug, getter `fullAddress.adress` is a `string` and is not an `Address`

https://github.com/exceljs/exceljs/blob/3e078cc4fae3c3a78214ccd3a75c91630e5ad08a/spec/unit/doc/cell.spec.js#L156

https://github.com/exceljs/exceljs/blob/master/lib/doc/cell.js#L22

https://github.com/exceljs/exceljs/blob/master/lib/doc/cell.js#L272

Siemienik:
I've just added next typing fixations (for Column interface)

PR #1093, opened 2020-01-27
Wait for writing to tmp fiels before handling zip stream close

sohai:
I'm having a hard time to write a failing test without those changes but in one commercial project, we have tests that were flaky without this.


sohai:
I moved resolving promise to finish listener of tempStream to handle a rare case when zip close event occurs before `entry.pipe(tempStream);` finish

alubbe:
@Siemienik @guyonroche I think we should merge this, but I'm not sure if we can write a test here. What do you think?

Siemienik:
This is not a perfect practice to commit something without tests. 

mhmm... could you test emitting error?  (Do you know any situation that should throw an exception?)

alubbe:
Maybe we could use proxyquire? I think the failure is actually really unlikely, the issue is actually around the first tmp callback potentially getting called AFTER the finished event has already been triggered, so `this.waitingWorkSheets` is empty even though it should have content

sohai:
@Siemienik I added one test. I totally understand that this is not good practice to commit something without tests but as I mentioned I could not figurated out how to do that properly, please let me know if you have some better ideas to test it.

PR #1095, opened 2020-01-29
Support ArrayBuffer as an xlsx.load argument

sohai:
And it looks like we need to treat it as StringChunk

You can check the problem here https://codesandbox.io/s/file-upload-in-react-7ibtx


onionhammer:
@alubbe not sure if this is related, but if you take this codesandbox sample, upgrade to latest exceljs and upload an xlsx file, the browser tab will lock up (100% CPU)

Issue #633, opened 2018-08-24
Existing Rich Text corrupts .xlsx on write


Hydro95:
My current program reads in an existing .xlsx that contains a few cells with values that look like this (this was made in excel and I haven't created the structure with JavaScript in any way):

```
{ richText:
   [ { font:
        { bold: true,
          size: 10,
          color: { theme: 1 },
          name: 'Arial',
          family: 2 },
       text: '(j) ' },
     { font: { size: 10, color: { theme: 1 }, name: 'Arial', family: 2 },
       text: 'blah blah blah ' },
     { font:
        { bold: true,
          underline: true,
          size: 10,
          color: { theme: 1 },
          name: 'Arial',
          family: 2 },
       text: 'blah blah' },
     { font: { size: 10, color: { theme: 1 }, name: 'Arial', family: 2 },
       text: ' more blah blah blah \r\n' },
     { font:
        { bold: true,
          size: 10,
          color: { theme: 4 },
          name: 'Arial',
          family: 2 },
       text: '\r\nReference: numbers, blah blah blah' } ] }
```

The problem is when I write the .xlsx back into a file using ``workbook.xlsx.writeFile`` and open in Excel the file is prompted to be repaired, clearing *only* the cells with rich text, and I get an error (in Excel) of: 
``Repaired Records: String properties from /xl/sharedStrings.xml part (Strings)``
``Removed Records: Named range from /xl/workbook.xml part (Workbook)``

A temporary solution I found was to just take the ``cell.value`` and set it equal to the ``cell.text`` which nukes the formatting while keeping the text, and everything works fine, but I was wondering if there was something wrong with the way rich text is being written back into a new file?

Thanks for any help or insight.

tony-rodriguez:
I am running into a similar issue. I am not seeing a corrupted file, but Rich Text values are lost on write. I have `{ useSharedStrings: true }`, but any cell that had existing Rich Text comes out with no value.

Kosalexei:
Before writing a file, you need to go through the existing cells with rich text and pass cell.model.value to cell.value:
```typescript
worksheet.getRow(3).eachCell(cell => {
  if (cell.model.value!.hasOwnProperty("richText")) {
    cell.value = cell.model.value as CellRichTextValue;
  }
});
```

PR #1099, opened 2020-01-30
Export shared strings with RichText

kigh-ota:
This PR is intended to fix https://github.com/exceljs/exceljs/issues/633.

When an Excel file is read, a shared string with rich text (`<r>`) nodes inside is correctly parsed and reconciled into a `CellXform` object like below.
```js
{
  model: {
    value: { richText: [ ... ] },
    type: Enums.ValueType.RichText,
  },
}
```
However, `XLSX#writeFile()` converts this object into an empty cell in the exported Excel file, because `CellXform#prepare()` and `#render()` do nothing when `model.type === Enums.ValueType.RichText`.

To fix this, the both methods now handle `RichText` in the same way as `String`.

### DIFF
--- a/lib/xlsx/xform/sheet/cell-xform.js
+++ b/lib/xlsx/xform/sheet/cell-xform.js
@@ -65,6 +65,7 @@ class CellXform extends BaseXform {
 
     switch (model.type) {
       case Enums.ValueType.String:
+      case Enums.ValueType.RichText:
         if (options.sharedStrings) {
           model.ssId = options.sharedStrings.add(model.value);
         }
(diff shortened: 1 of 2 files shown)


Siemienik:
hmm.. should we add some documentation about these changes?

kigh-ota:
@Siemienik 
Which documentation do you mean, for users or for developers?
In my opinion, one for users is not mandatory.
This is because the documentation already includes an example of a rich text value with multiple styles, so users (including me) would expect that the value would be exported as is even without additional description.
https://github.com/exceljs/exceljs/blob/a2eaa9e75e477a9f5c88d54979bd95eb37571c3d/README.md#rich-text

But I'm new here, so tell me if I should consider more :)

Siemienik:
@kigh-ota ok, it's my gap, sorry abut this, You are right :)

kigh-ota:
@Siemienik
Thank you for your comments!

Issue #623, opened 2018-08-02
Issue with borders for merged cell when rewriting an excel workbook


kami23:
I am using this function to create a copy of my excel file named mytest.xlsx
i have merged cells with borders inside it but in the new file (the copy) some of the merged cells lost their borders i don't know why?  

function createExce()
    {
       workbook.xlsx.readFile('mytest.xlsx')
       .then(function(workbook) {
        // use workbook
return workbook.xlsx.writeFile('new.xlsx');
    });

    }

jeremyjh:
@kami23 did you find a workaround for this?

kami23:
@jeremyjh no i didn't find a solution for it i had to use a loop to draw all the borders

Toub:
I reproduce a similar issue with the following code:

```ts
 const c1 = worksheet.getCell(2, 2);
  c1.border = {
    top: {
      style: 'medium',
      color: { argb: 'FF333333' },
    },
    left: {
      style: 'medium',
      color: { argb: 'FF333333' },
    },
    bottom: {
      style: 'medium',
      color: { argb: 'FF333333' },
    },
    right: {
      style: 'medium',
      color: { argb: 'FF333333' },
    },
  };
  const c2 = worksheet.getCell(3, 3);
  c2.merge(c1);
```
Displayed result with Excel 2016:

![image](https://user-images.githubusercontent.com/367762/72867940-9272bf80-3cf1-11ea-9981-4c7e7930239c.png)

Displayed result with Libre Office 6.1:

![image](https://user-images.githubusercontent.com/367762/72868098-12008e80-3cf2-11ea-82e5-2b7bacac1355.png)

npmDude:
The reason I think this is happening is because merged cells link their styles from the cell.master which is the top left cell of the merged cells. So when you load a file that has a merged cell it will link its style from the master which only has the top and left borders hence merged cells only has top and left borders.

The workaround for this one is to set the borders of the master again.

npmDude:
The reason this is happening is because when ExcelJS is reading the file it will read the merged cell and link its style based on the master cell (top-left cell). When you merge cells the border style of the master cell will only be on the top and left borders that is why that is the only ones left later on.

But after reading the file, it seems that the link on the style of merged cells is gone that is why when you set the borders for the master again, it is not applying to the other ones.

The workaround for this one is to unmerge the cells programmatically, set the borders of the master cell and merge the cells again.

npmDude:
[...]

Issue #635, opened 2018-08-28
The merged cells do not have enough border information


zengwe:
For example, I merged the cells $B$2:$C$3 and set the outer border to red. I get each cell through EXCELJS, with only the information for the top and left borders.
Then I merged $B$5:$C$6 and set only the top and left borders. The resulting border information is the same as $B$2:$C$3.
So how do I know if the right and bottom borders have color settings?

dogusev:
It's an issue, I also faced with it.
It seems that border styles are copied from the master cell, and ignore all others from the merged range

cbeltrangomez84:
Is the same #623

Toub:
Not sure if it is exactly the same problem, but I provided a simple reproduction code on #623

Siemienik:
hmm, it probably happens when you read excel files. 
for a workaround, You may use `cell.master` 
https://github.com/exceljs/exceljs/blob/master/lib/doc/cell.js#L188

When merging cells has been proceded from code it should work ok:
https://github.com/exceljs/exceljs/blob/master/lib/doc/cell.js#L169

I think it's a good topic for pull request if anyone wants :)

npmDude:
The reason I think this is happening is because merged cells link their styles from the cell.master which is the top left cell of the merged cells. So when you load a file that has a merged cell it will link its style from the master which only has the top and left borders hence merged cells only has top and left borders. 

The workaround for this one is to set the borders of the master again.

npmDude:
The reason this is happening is because when ExcelJS is reading the file it will read the merged cell and link its style based on the master cell (top-left cell). When you merge cells the border style of the master cell will only be on the top and left borders that is why that is the only ones left later on.

But after reading the file, it seems that the link on the style of merged cells is gone that is why when you set the borders for the master again, it is not applying to the other ones.

The workaround for this one is to unmerge the cells programmatically, set the borders of the master cell and merge the cells again.

PR #1102, opened 2020-01-31
Keep borders of merged cells after rewriting an Excel workbook

kigh-ota:
This PR is intended to fix https://github.com/exceljs/exceljs/issues/623.

# Cause of the bug

Currently, styles of cells included in a merged cell are merged and always the same after reading from file, because `Worksheet#set model()` calls `Cell#merge()`, which overrides the cell's style with the master (left-top) cell.
https://github.com/exceljs/exceljs/blob/cb290926adca692402a722b7c577b1fa36a6b953/lib/doc/worksheet.js#L755
https://github.com/exceljs/exceljs/blob/cb290926adca692402a722b7c577b1fa36a6b953/lib/doc/cell.js#L172

However, in the original xml, each cell in a merged cell may have different styles.
For example, [this file](https://github.com/exceljs/exceljs/pull/1102/files#diff-bf5614c9fbce89800bf09bf72d456894), which I created with Excel, has `sheet1.xml` like this:
```xml
  <sheetData>
    <row r="2" spans="2:3">
      <c r="B2" s="1"/>
      <c r="C2" s="2"/>
    </row>
    <row r="3" spans="2:3">
      <c r="B3" s="3"/>
      <c r="C3" s="4"/>
    </row>
  </sheetData>
  <mergeCells count="1">
    <mergeCell ref="B2:C3"/>
  </mergeCells>
```
Thus, merging the styles drops intended cell styles, which causes https://github.com/exceljs/exceljs/issues/623.

# Solution

`Worksheet#set model()` is changed such that it parses merged cells without overriding the styles of included cells.

## Note

I keep the behavior of `Worksheet#mergeCells()` unchanged, since this is an API that can be used programmatically. https://github.com/exceljs/exceljs#merged-cells
Another related issue https://github.com/exceljs/exceljs/issues/635 seems to correspond to this programmatic use case, but I think it can be tackled separately.


npmDude:
Can I know when is the scheduled merge for this PR?

Siemienik:
in the time when @guyonroche  has a minute :)

PR #1104, opened 2020-02-03
Fix #1103: `editAs` not working

### LINKED ISSUE #1103: `editAs` does not work when adding image over a range
set `editAs` to `absolute`, but it is still `oneCell`.
```javascript
ws.addImage(imageId, {
  tl: { col: 0.1125, row: 0.4 },
  br: { col: 2.101046875, row: 3.4 },
  editAs: 'absolute'
});
```

Alanscut:
## Changes:
* fix for issue #1103


alubbe:
I'm not entirely sure what this PR is about - @Siemienik do you know? @Alanscut could you elaborate a little bit and maybe provide a test case?

Alanscut:
ok, I will add some test for it later.
when you add an image to a worksheet, property `editAs` will control how the image is anchored to the cell(s). it could be undefined, `oneCell` or `absolute`. But when you set it to `absolute`, the value of `editAs` in generated excel file will still be `oneCell`.
https://github.com/exceljs/exceljs/blob/master/README.md#add-image-over-a-range
pr-1104

spec/integration/workbook/images.spec.js
```diff
@@ -218,7 +218,7 @@ describe('Workbook', () => {
           expect(images.length).to.equal(1);
 
           const imageDesc = images[0];
-          expect(imageDesc.range.editAs).to.equal('oneCell');
+          expect(imageDesc.range.editAs).to.equal('absolute');
```
Alanscut: @alubbe  if not fixed, the actual value will be `oneCell`.
Since there are already many tests for `oneCell`, I am directly modifying my previous test cases.

Alanscut:
there is another problem. The README says `undefined` is the default value of  property `editAs`, however, the actual situation is that `oneCell` is the default value. In fact, when creating a Excel file and inserting an image to it, the default property of editAs is `oneCell` too. So should I change the readme file?

spec/integration/workbook/images.spec.js
Siemienik: ok, this should cover these changes :)

have we any test for "oneCell" that makes exactly same thing like this one?

Siemienik: > have we any test for "oneCell" that makes exactly same thing like this one?

ok, it is:)

Siemienik:
AD. `undefined` in documentation - I think, we should update readme, and all will be ok.

alubbe:
@Alanscut yeah let's update the README and then we're good to go

PR #1107, opened 2020-02-04
fix some errors and typos in readme

Alanscut:
## Changes:
* `showGridLines` is not in `properties`, but in `pageSetup` and `views`
```javascript
// showGridLines will not work
var sheet = workbook.addWorksheet('My Sheet', {properties: {showGridLines: false}});
```
it should be set like this:
```
// create a sheet where the grid lines are hidden when opening it
var sheet = workbook.addWorksheet('My Sheet', {views: [{showGridLines: false}]});
```
or
```
// create a sheet where the grid lines are hidden when printing it
var sheet = workbook.addWorksheet('My Sheet', {pageSetup: {showGridLines: false}});
```

* some typos


README.md
```diff
@@ -261,7 +261,7 @@ For Example:
 var sheet = workbook.addWorksheet('My Sheet', {properties:{tabColor:{argb:'FFC0000'}}});
 
 // create a sheet where the grid lines are hidden
-var sheet = workbook.addWorksheet('My Sheet', {properties: {showGridLines: false}});
+var sheet = workbook.addWorksheet('My Sheet', {views: [{showGridLines: false}]});
```
Siemienik: I confirm these changes: https://github.com/exceljs/exceljs/blob/master/index.d.ts#L1219

PR #1112, opened 2020-02-04
Update issue templates

Siemienik:
We have a lot of bugs reported which has a poor description, it does not help us to reproduce and fix it quickly.
So I think the issue template should improve the situation :)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,70 +1,35 @@
 ---
-name: Bug report
+name: 🐛 Bug report
 about: Create a report to help us improve
 title: "[BUG] XYZ"
 labels: bug
-assignees: ''
-
 ---
 
-<!--
-Bug description + screenshots
--->
+## 🐛 Bug Report
 
-Lib version:
+<!-- A clear and concise description of what the bug is. -->
+
+Lib version: X.Y.Z
 
 ## Steps To Reproduce
 
-1.
-2.
-3.
-
-### code example:
+<!-- The exact steps required to reproduce the issue, ideally with a code example -->
 
 ```javascript
-/* please paste code here */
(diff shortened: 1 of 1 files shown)


alubbe:
I think that's a great idea, but I believe it needs to be short - otherwise users might end up ignoring it. Personally, I like https://raw.githubusercontent.com/facebook/jest/master/.github/ISSUE_TEMPLATE/bug.md so I've shortened our template to be similar to that. What do you think?

Also pinging @guyonroche

Alanscut:
why not also add a PR template ?😃

Siemienik:
@Alanscut yeach it;s great idea:)

Siemienik:
@alubbe for me the important thing is to give me a test which will fail before fixation. We have a lot of bugs reported, that we need a lot of time to inspect and analyze them.

Test code for an issue will give us:

set the quality of issues to the highest level
reduce the fixation time cost
encourage to creating new PRs (if someone makes a test it will probably make a fixation :) )
In other side, You have right, the short template should increase bug reports amount.
So I have shared my thoughts with you, and I'm really open to change my vision to something else. 

@alubbe @Alanscut what do you think?

alubbe:
I completely agree, that's why I tried to find a 'middle-ground' approach, that both looks short, so that it actually gets read instead of deleted, but still wants a code example. That's not as much as a fully blown test case, but it's close enough. Here's my suggestion:
```
## Steps To Reproduce

<!-- The exact steps required to reproduce the issue, ideally with a code example and an expectation (like in a unit test) -->

```javascript
const wb = new ExcelJS.Workbook();
const ws = wb.addWorksheet('XYZ');
ws.getCell('A1').value = 7;
expect(ws.getCell('A1').value).to.equal(7);
```

alubbe:
@Alanscut this PR adds more templates: https://github.com/exceljs/exceljs/pull/1128

PR #1125, opened 2020-02-13
Replace uglify with terser

alubbe:
Uglify can not handle ES6+, such as `const` and `let`. Some npm dependencies, like `saxes`, can not be used without replacing uglify with the more modern terser because they publish modern JS.

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -3,7 +3,7 @@
 module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-babel');
   grunt.loadNpmTasks('grunt-browserify');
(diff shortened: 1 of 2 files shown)


Siemienik:
@alubbe 
1. something went wrong while testing on `node:8`, could you check what's going on?
2. hmm.. currently I haven't a possibility to check these changes but for me, that's really ok 😄

alubbe:
I don't understand the failure here - because it looks like mocha crashes, for no clear reason, but this PR https://github.com/exceljs/exceljs/pull/1127 which has the same change, is somehow green. Is it flaky? Am I missing something?

alubbe:
I wiped the cache and now it's green... :D

PR #1126, opened 2020-02-13
Apply codestyles on each commit and run lint:fix

alubbe:
Since we started using prettier and eslint to maintain our code styling, we've started diverging because it's not automatically applied. This PR adds prettier to the pre-commit hook as I use it on my other projects and runs `lint:fix` on the entire code base to level the playing field for other PRs and avoid their changelogs containing these style fixes.


Siemienik:
that's really great idea, but not all rules are fine for me.
For example this one:
https://github.com/exceljs/exceljs/pull/1126/files#diff-da27af84b9fac78a1da037b11fe9ac47L36

These changes probably make lots of conflicts in the nearest future.

maybe we should redesign linter rules? 

@alubbe @guyonroche what do you think?

alubbe:
You mean the formatting of the `switch` statement? I'm pretty sure it's an opionated setting of `prettier` that cannot be changed. Also, there are almost no PRs open right now, so if we merge soon, it shouldn't affect anyone's work, I hope. Even if it does, they just need to run `lint:staged` on their changeset and then their PR will be probably be fixed - or I help by hand, like last time. But this time, it should be the last time because going forward every PR will have the right formatting.

Siemienik:
I'v writen about this one:
image

but inside the switch statement is a similar situation :)

For me, one -line statements are very convenient but it isn't a big deal for me to keep new conventions

alubbe:
Ah gotcha. But yeah, same situation: prettier has an opinion and won't let you change it ;)
What's good about this choice is that it's easier to extend what each of these functions (or case blocks) do, and reduces the diffs on future PRs when we review those changes

alubbe:
I'll merge this for now, so that at least our code base reflects what our prettier+eslint rules specify - if we don't like our rules, let's create a new issue or PR to update them.

Issue #748, opened 2019-02-19
Backlog


Siemienik:
Hi, I think we should update our backlog and add:
1. Performance
2. clean up issues

IgnatovDan:
I would prefer Performance: generate [10000 rows x 50 cols] xlsx file in 3 seconds ([see conversation](https://github.com/exceljs/exceljs/issues/742))

alubbe:
I like those three. Right now I'm quite interested in improving the streaming, esp. adding cancellation, can we add that as a fourth?

alubbe:
As a fifth point, I also think we should explore replacing sax (unmaintained for over 2 years) with saxes, which also speeds up our runtime performance (at least for streaming). And I believe @coldhiber also found an issue with the temp file library that uses global state

Siemienik:
1. Performance
2. Clean up issues
3. Do fixes
4. Streaming, esp. adding cancellation
5. Replacing sax  with saxes

alubbe:
I feel we're making good progress on 1, 4 and 5 - @Siemienik what's your take on 2 & 3?

Siemienik:
ad2. a lot of work what I did gives a result that around 30% current opened issues has labels,
for instance: bug labels: 101 unresolved today

now 427 are open from all 869 issues, I was involved in 192 closed issues.

lots of works we have to do to clear it up

ad3. It relevant enjoyable feeling when I see amazing numbers of a pull request that our fantastic community done. So I am in the best mind for the future :smile:


That I see for now. We should update this list by adding:
6. better typscript integration (some types aren't consistent) 
7. add tests and examples with Angular(some devs has issue with that)

PR #1127, opened 2020-02-13
[WIP] Replace sax with saxes

alubbe:
THIS PR IS NOT READY TO BE MERGED - IT IS A PoC

As mentioned last year (https://github.com/exceljs/exceljs/issues/748#issuecomment-475109436), I believe we should consider replacing `sax` with `saxes`. Here are my main motivating reasons:

1. `sax` has not been maitained in close to 3 years
2. `saxes` uses ES6 and modern streams, which should enable us to use async iterators, which a) massively simplifies the code responsible for the stream control logic and b) handles backpressure for us, which could mean we might be able to drop our custom flow control implementation
3. `saxes` is faster than `sax`

This PR is a work in progress right now and cannot be merged until `saxes` v5 is officially released. It also requires that we replace uglify with terser to handle the ES6 code that `saxes` publishes.

Looking forward to your feedback!


alubbe:
This PR had to do a number of things, so I've spun out 3 separete PRs out of it. Once they are merged, the diff will be smaller and easier to read.

https://github.com/exceljs/exceljs/pull/1125
https://github.com/exceljs/exceljs/pull/1126
https://github.com/exceljs/exceljs/pull/1124

PR #1128, opened 2020-02-13
Add PR, Feature Request and Question github templates

alubbe:
Building on https://github.com/exceljs/exceljs/pull/1112, here are my suggestions for the PR, Feature Request and Question github templates.

Again, they are heavily inspired by https://github.com/facebook/jest/tree/master/.github - but there's one big difference: They don't want the github issue tracker to be used for questions, which are instead handled via discord, stackoverflow, etc. What is our stance here? 

We could encourage users to open issues to ask questions but warn them that they might not get answered (but then where should they go). Or we could ask them to use this only for clear bug reports or feature requests and direct them elsewhere (but where). What do you think?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,8 +1,8 @@
 ---
 name: 🐛 Bug report
 about: Create a report to help us improve
-title: "[BUG] XYZ"
-labels: bug
+title: '[BUG] XYZ'
+labels: ':bug: Bug'
 ---
 
 ## 🐛 Bug Report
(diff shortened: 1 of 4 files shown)


Siemienik:
I use only Github, so ... :)

I think that we could encourage users to open issues to ask questions without warns 🥇
I attempt to respond to 5-10 issue per day, add labels, close or - that I mostly like - resolve them.

Fixing the bugs is for me the important thing. But very often I haven't enough time to do that. So I'm really happy and thankful when someone from our community does it 🚀 🚀 🚀

alubbe:
@Siemienik I've tried to describe our approach in the template - what do you think? 

```
Have a question about exceljs? Need help using it a certain way that our README does not seem to explain? Ask away!

Please keep in mind that we're unpaid open souce volunteers, so it will help us greatly if you add as many details as possible to your question and describe what other approaches you've tried out so far. And of course, we'd love for you to contribute back the answer to our README via a pull request!
```

Issue #997, opened 2019-10-25
ERROR in node_modules/exceljs/index.d.ts(1661,34): error TS2503: Cannot find namespace 'NodeJS'.


aakashjangid:
Code is failed to compile and giving below error - 
ERROR in node_modules/exceljs/index.d.ts(1661,34): error TS2503: Cannot find namespace 'NodeJS'.

Details - 
exceljs - ^3.3.0
@angular/cli - 7.3.9

I already tried each and every suggestion from everyone, but nothing worked for me.

Need this error to be removed so that code gets compiled successfully.

langkydoanh:
The same issue:
ERROR in node_modules/exceljs/index.d.ts(1661,34): error TS2503: Cannot find namespace 'NodeJS'.
"@angular/cli": "~7.0.4", 
"@types/node": "~8.9.4",
"exceljs": "^3.3.0",

srivishalsangepu:
Failed to compile.

node_modules/exceljs/index.d.ts(1661,34): error TS2503: Cannot find namespace 'NodeJS'.

agamgupta6:
Same error for me 

"@angular/animations": "~7.2.0",
    "@angular/common": "~7.2.0",
    "@angular/compiler": "~7.2.0",
    "@angular/core": "~7.2.0",
    "@angular/forms": "~7.2.0",
    "@angular/platform-browser": "~7.2.0",
    "@angular/platform-browser-dynamic": "~7.2.0",
    "@angular/router": "~7.2.0",
    "core-js": "^2.5.4",
    "exceljs": "^3.3.0",
    "file-saver": "^2.0.2",

smithtaylorg:
I am getting the same error: ERROR in node_modules/exceljs/index.d.ts(1661,34): error TS2503: Cannot find namespace 'NodeJS'. I tried the suggestions in #971, but none of those worked for me. Anyone find a fix yet?  
"exceljs": "^3.3.1"
"core-js": "^2.6.10"
"@angular-devkit/build-angular": "^0.13.9",
"@angular/animations": "^7.2.15",
"@angular/cli": "^7.3.9",
"@angular/common": "^7.2.15",
@angular/compiler": "^7.2.15",
"@angular/compiler-cli": "^7.2.15",
"@angular/core": "^7.2.15",
"@angular/forms": "^7.2.15",
"@angular/http": "^7.2.15",
"@angular/language-service": "^7.2.15",
"@angular/platform-browser": "^7.2.15",
[...]

Issue #971, opened 2019-09-20
ERROR in node_modules/exceljs/index.d.ts(1648,34): error TS2503: Cannot find namespace 'NodeJS'.


phanbien9x:
My application can not start after integrated exceljs. I use angular 7. I checked on node_modules/exceljs/index.d.ts but there are not error. I tried to run example with angular 6 then it work. Please let me know how to implement exceljs in angular 7. Advanced thanks!

I use:
"exceljs": "^2.0.1",
"file-saver": "^2.0.2",

"@angular/cli": "^7.3.0",
"@types/node": "^12.7.5",

sasapet:
I have the same issue. 
The other import 
import * as Excel from "exceljs/dist/exceljs.min.js";
Fixes the issue, but then you don't get the types.

guiqui:
Change the version of @types/node to "@types/node": "^10.12.0" works for me.

sanjayrane:
I have the same issue.
The other import
add  tsconfig.json  below code
"compilerOptions" :{
"exceljs": [
        "node_modules/exceljs/dist/exceljs.min.js"
      ]
}

phanbien9x:
I tried all of your suggestion but it only work when I use "workbook" imported by @sasapet way like this:
const workbook = new Excel.Workbook();
Thank all of you to help me! :)

neeravpanchal26:
Instead of importing
Use it in the following way in your function: 
**const Excel = require('exceljs');**

Example:
// Excel exporter
    ExportToExcel(Header, ReportName): void {
        // Excel object
        const Excel = require('exceljs');

        //  Create workbook
        const workbook = new Excel.Workbook();

        // Create worksheet
        const worksheet = workbook.addWorksheet(ReportName);

        // Title row
        const titleRow = worksheet.addRow([ReportName]);
        // Customize Title row
        titleRow.font = { name: 'Helvetica Neue', family: 4, size: 16};

        // Export to file
        workbook.xlsx.writeBuffer().then((data) => {
            FService.saveAs(new Blob([data]), ReportName + '_export_' + new Date().getTime() + '.xlsx');
        });
    }

GustavDjavulen:
I erase the NodeJS.TypedArray from "index.d.ts" in de "node_modules/exceljs"

that works for me :)

![image](https://user-images.githubusercontent.com/54324708/69768198-45d19480-1145-11ea-8fb1-f8ebb7356694.png)

not the best form to fix it but works

[...]

PR #1133, opened 2020-02-17
Update @types/node version to latest lts

Siemienik:
This one should fix issues #997 and #971


desarrollopetrocasinos:
Hi. I have "@types/node": "^13.13.51", and i still getting that error issue #997

PR #1135, opened 2020-02-18
[MAJOR VERSION] Async iterators

alubbe:
This PR is the result of a lot of groundwork laid by previous PRs https://github.com/exceljs/exceljs/pull/1127 https://github.com/exceljs/exceljs/pull/1125 https://github.com/exceljs/exceljs/pull/829 https://github.com/exceljs/exceljs/pull/1142 https://github.com/exceljs/exceljs/pull/1190 https://github.com/exceljs/exceljs/pull/1140 https://github.com/exceljs/exceljs/pull/1139 https://github.com/exceljs/exceljs/pull/1142

I'm happy to report that this PR adds 200 lines of documentation and deletes over 400 lines of code while keeping all tests green, maintaining performance and at the same time making our code base much easier to reason about and maintain. 

I've thrown out everything related to streams & custom flow control logic around that and replaced it with generators (sync wherever possible, async when necessary). I exposed one external stream API at the end, built on the generators underneath.

This PR is now ready to be merged. It supports node v8 via the es5 import.

Here is the new usage:

```js
const workbookReader = new ExcelJS.stream.xlsx.WorkbookReader('./file.xlsx');
for await (const worksheetReader of workbookReader) {
  for await (const row of worksheetReader) {
    // ...
  }
}
```


alubbe:
All the tests are green - I'm really excited what you think of this approach @Siemienik @guyonroche 

If we like it, I would love to add support for it our streaming reader, as well, so that the consumers of our library can use for-await-of on the streams that we expose

... and make our streams destroyable without creating memory leaks ;)

alubbe:
And here are the benchmarking numbers. First, current `master`:
```
####################################################
WARMUP: Current memory usage: 8.63 MB
WARMUP: huge xlsx file profiling started
Reading worksheet 1
Reading row 50000
Reading row 100000
Reading worksheet 2
Reading row 150000
Processed 2 worksheets and 150002 rows
WARMUP: huge xlsx file profiling finished in 6961ms
WARMUP: Current memory usage (before GC): 134.96 MB
WARMUP: Current memory usage (after GC): 51.7 MB

####################################################
RUN 1: huge xlsx file profiling started
Reading worksheet 1
Reading row 50000
Reading row 100000
Reading worksheet 2
Reading row 150000
Processed 2 worksheets and 150002 rows
RUN 1: huge xlsx file profiling finished in 6490ms
RUN 1: Current memory usage (before GC): 177.75 MB
RUN 1: Current memory usage (after GC): 51.78 MB

####################################################
RUN 2: huge xlsx file profiling started
Reading worksheet 1
Reading row 50000
Reading row 100000
Reading worksheet 2
Reading row 150000
Processed 2 worksheets and 150002 rows
RUN 2: huge xlsx file profiling finished in 6683ms
RUN 2: Current memory usage (before GC): 148.7 MB
RUN 2: Current memory usage (after GC): 51.67 MB

####################################################
RUN 3: huge xlsx file profiling started
Reading worksheet 1
Reading row 50000
Reading row 100000
Reading worksheet 2
Reading row 150000
Processed 2 worksheets and 150002 rows
RUN 3: huge xlsx file profiling finished in 6350ms
RUN 3: Current memory usage (before GC): 176.76 MB
RUN 3: Current memory usage (after GC): 51.63 MB
```
And after this PR:
```
####################################################
WARMUP: Current memory usage: 8.61 MB
WARMUP: huge xlsx file profiling started
Reading worksheet 1
Reading row 50000
Reading row 100000
Reading worksheet 2
Reading row 150000
Processed 2 worksheets and 150002 rows
WARMUP: huge xlsx file profiling finished in 6823ms
WARMUP: Current memory usage (before GC): 148.66 MB
WARMUP: Current memory usage (after GC): 41.22 MB

####################################################
RUN 1: huge xlsx file profiling started
Reading worksheet 1
Reading row 50000
Reading row 100000
Reading worksheet 2
Reading row 150000
Processed 2 worksheets and 150002 rows
RUN 1: huge xlsx file profiling finished in 6749ms
RUN 1: Current memory usage (before GC): 115.94 MB
RUN 1: Current memory usage (after GC): 25.15 MB

####################################################
RUN 2: huge xlsx file profiling started
Reading worksheet 1
Reading row 50000
Reading row 100000
Reading worksheet 2
Reading row 150000
Processed 2 worksheets and 150002 rows
RUN 2: huge xlsx file profiling finished in 6653ms
RUN 2: Current memory usage (before GC): 97.41 MB
RUN 2: Current memory usage (after GC): 25.12 MB

####################################################
RUN 3: huge xlsx file profiling started
Reading worksheet 1
Reading row 50000
Reading row 100000
Reading worksheet 2
Reading row 150000
Processed 2 worksheets and 150002 rows
RUN 3: huge xlsx file profiling finished in 6797ms
RUN 3: Current memory usage (before GC): 145.61 MB
RUN 3: Current memory usage (after GC): 41.11 MB
```
In short, we have a slight slowdown (3-4%), but lower memory usage and much simpler control flow in our code, making it more maintainable.

alubbe:
Okay I'm quite happy with the state of this PR now - over 400 lines of code deleted while keeping all tests green, maintaining performance and at the same time making our code base much easier to reason about and maintain. I will start testing this branch on a larger application, but it is now ready for review and I'm really looking forward to your feedback!

In short, the trick is to throw out everything related to streams and custom flow control logic around that an to replace that with generators (sync wherever possible, async when necessary) and then to expose one external stream API at the end, built on the generators underneath.

alubbe:
Async iteration example:

```js
const stream = fs.createReadStream(...);
const wb = new ExcelJS.stream.xlsx.WorkbookReader();
const options = {
  entries: 'emit',
  sharedStrings: 'cache',
  worksheets: 'emit',
};

for await (const {eventType, value, entry} of wb.parse(stream, options)) {
  if (eventType === 'worksheet') {
    worksheetCount += 1;
    for await (const events of value.parse(entry, options)) {
      for (const event of events) {
        if (event.eventType === 'row') {
          ...
        }
      }
    }
  }
}
```

Streaming example:

```js
const stream = fs.createReadStream(path.join(__dirname, 'spec/integration/data/huge.xlsx'));
const wb = new ExcelJS.stream.xlsx.WorkbookReader();
const options = {
  entries: 'emit',
  sharedStrings: 'cache',
  worksheets: 'emit',
};
wb.read(stream, options);

wb.on('error', (err) => {
  ...
});
wb.on('worksheet', worksheet => {
  worksheet.on('row', () => {
    ...
  });
});
wb.on('end', () => {
  ...
});
```

alubbe:
And just to re-iterate, this is still 100% compatible with node 8, it just requires our library consumers to change their import to the ES5 version. What do you think?

Siemienik:
I'm going to review it soon

alubbe:
Thanks to the feedback of @Siemienik I have built some default options and added a way of consuming everything in a really simple way:
```js
const workbook = new ExcelJS.stream.xlsx.WorkbookReader('./file.xlsx');
for await (const worksheetReader of workbookReader) {
  for await (const rows of worksheetReader) {
    for (const row of rows) {
      // ...
    }
  }
}
```
This is really awesome now :D
Please note that `worksheetReader` returns an array of rows rather than each row individually for performance reasons: https://github.com/nodejs/node/issues/31979

alubbe:
@Siemienik I've added some lines to index.d.ts, but I have no experience with TypeScript - could you please check whether I got it right?

Siemienik:
@alubbe ok, I will review it tomorrow. Great job Andreas:)

Siemienik:
@guyonroche @alubbe What are you thinking about making external libraries: `exceljs/workbook-reader`, `exceljs/workbook-writer`? (and move some code there)

alubbe:
Yeah I think that would make a lot of sense - but I won't have the time for the next weeks and months to work on that. So how about we keep in inside exceljs for now and extract it later?

lib/stream/xlsx/worksheet-reader.js
```diff
@@ -87,14 +89,35 @@ class WorksheetReader extends EventEmitter {
...
+  }
 
-  _emitRow(row) {
-    this.emit('row', row);
+  async *[Symbol.asyncIterator]() {
+    for await (const events of this.parse()) {
+      const rows = [];
+      for (const {eventType, value} of events) {
+        if (eventType === 'row') {
+          rows.push(value);
```
Siemienik, line 110: hmm... will it has performance effort if here you put:
```javascript
    yield value;
```

It should make these simpler:
```javascript
for await (const worksheetReader of workbookReader) {
  for await (const row of worksheetReader) {
//without this:    for (const row of rows) {
      // ...
    //}
  }
}
```

as you mention, that should be slower, but if could I ask you to check it again and show me benchmark result?
I a little bit worry about backward compatibility it the future when we would keep this API.

spec/integration/pr/test-pr-1093.spec.js
```diff
@@ -1,33 +0,0 @@
-const proxyquire = require('proxyquire');
```
Siemienik, line 1: why these files were deleted?

Siemienik:
@Alanscut  would you review these changes too? I think that You know this library really well, so your opinion will worth a lot for me.

index.d.ts
```diff
@@ -1781,5 +1776,53 @@ export namespace stream {
 			addWorkbookRels(): Promise<void>;
 			addWorkbook(): Promise<void>;
 		}
+
+
+		interface WorkbookStreamReaderOptions {
+			/**
+			 * @default 'emit'
+			 */
+			entries: 'emit' | void;
```
Siemienik, line 1785: I don't understand whats means default here? how do you determine if it's set or not?

Siemienik, line 1785: TS has two ways to define optional parameters like here:
```typescript
			first: 'emit' | 'cache' | void;
//or
			second?: 'emit' |  'cache';
```
To be honest, I will prefer the second way (with a question mark).

The most perfect way is use separate types for each property:
https://github.com/exceljs/exceljs/pull/1199/files#diff-b52768974e6bc0faccb7d4b75b162c99R995

spec/integration/pr/test-pr-1093.spec.js
alubbe: this test relies on `proxyquire`, which messes a lot with `verquire`, making some of our tests flaky. I was part of the PR that added this test and can confirm that the thing it tests has become mostly irrelevant with the new control flow. Therefore, it's fine to delete this test and increase the overall test suite stability - but maybe @sohai can confirm?

index.d.ts
```diff
@@ -1289,11 +1289,6 @@ export interface Xlsx {
 	 */
 	load(buffer: Buffer): Promise<Workbook>;
 
-	/**
-	 * Create input stream for reading
-	 */
-	createInputStream(): import('events').EventEmitter;
```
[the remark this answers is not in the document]
alubbe: it's been removed, see reasons above

README.md
```diff
@@ -1896,9 +1898,12 @@ workbook.xlsx.readFile(filename)
     // use workbook
   });
 
-// pipe from stream
+// read from stream
 var workbook = new Excel.Workbook();
-stream.pipe(workbook.xlsx.createInputStream());
```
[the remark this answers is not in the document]
alubbe: No, it's been removed because I removed `.createInputStream()`.  The reason is that a) it can be completely replaced by `.read()` and b) it uses a lot of non-standard, custom code and behaves in really unexpected ways.

Since there is a replacement for it (`.read()`), we don't lose any functionality, but we make it   easier to maintain the library. However, some consumers of our library will have to make (simple) changes in order to use `4.0.0`, so is it worth it? Let's talk about the second point in detail:

It looks nice that we can use `stream.pipe(createInputStream())`, but if you look at the implementation if `createInputStream`, you start to realize that it's not a stream at all (https://github.com/exceljs/exceljs/pull/1135/files#diff-bae1d9f34b0ea7278956352e70c79c5e). It's a super custom implementation based on event emitters, has nothing do with actual writeable streams and re-implements a lot of logic itself. The code itself even says:
```
// The purpose of this module is to wrap the js-zip library into a streaming zip library
// since most of the exceljs code uses streams.
// One day I might find (or build) a properly streaming browser safe zip lib
```
The good thing is that with generators, we can actually build something that is standard, simple and easy to maintain, like I did here: https://github.com/exceljs/exceljs/pull/1135/files#diff-582ff38cb0a89a7a808debcaf9965d00R260

Now that the internals have been drastically simplified using the simpler async iteration control flow, it's actually difficult to re-expose a function that behaves just like `createInputStream` - and coupled with the fact that you could just use `.read()`, I think we should drop it.

index.d.ts
alubbe: I've just pushed an update to that - does that make it clearer?

spec/integration/pr/test-pr-1093.spec.js
sohai: yep, that makes sense to delete those tests now

README.md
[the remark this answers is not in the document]
AdrianSkierniewski: 👍 for getting rid of flow control

sohai:
that looks really cool, I hope this can be merged soon

AdrianSkierniewski:
good job 👍

alubbe:
@Siemienik thanks for pushing on simplifying the API even further. I've just run a series of benchmarks and I found that if we change `parse()` to emit each event instead of an array of events, we lose about 20% performance, so I would suggest to keep that. However, we can change the async iterator implementation based on `parse()` to emit each `row`, which results in a 5% slowdown, but gives us a much nicer interface:
```js
const workbook = new ExcelJS.stream.xlsx.WorkbookReader('./file.xlsx');
for await (const worksheetReader of workbookReader) {
  for await (const row of worksheetReader) {
    // ...
  }
}
```

lib/stream/xlsx/workbook-reader.js
```diff
@@ -20,11 +20,12 @@ class WorkbookReader extends EventEmitter {
 
     this.input = input;
 
-    this.options = options || {
-      hyperlinks: 'cache',
-      sharedStrings: 'cache',
-      worksheets: 'emit',
-    };
+    this.options = options || {};
```
Siemienik, line 23: I think this default options handling method is ok, I've just wanted to share with you about what I think is perfect:

```js
const defaults = Object.freeze({a:2,b:3,c:4}); //may be const for class

function someFunc(options){
  options = Object.freeze(Object.assign({},defaults,options))
  // ...
  console.log(options); //options is immutable
};

someFunc({a:1,c:5});
someFunc();
someFunc("asd");
someFunc(123);
someFunc([123]);
someFunc(null);
someFunc({});
```

image

it's only for info, I don't need changes here 😄

index.d.ts
Siemienik: yeah, that's perfect

README.md
```diff
@@ -2207,6 +2213,90 @@ workbook.commit()
...
+The streaming XLSX workbook reader is available in the ExcelJS.stream.xlsx namespace.
+
+The constructor takes a required input argument and an optional options argument:
+
+| Argument              | Description |
+| --------------------- | ----------- |
+| input (required)      | Specifies the name of the file or the readable stream from which to read the XLSX workbook. |
+| options (optional)    | Specifies how to handle the event types occuring during the read parsing. |
+| options.entries       | Specifies whether to emit entries (`'emit'`) or not (`undefined`). Default is `'emit'`. |
+| options.sharedStrings | Specifies whether to cache shared strings (`'cache'`), emit them (`'emit'`) or skip them (`undefined`). Default is `'cache'`. |
```
Siemienik, line 2227: may I please you to give some more explanations what's mean cache? 
Is it keep in the worksheet model, and it's available to read later, isn't it?

README.md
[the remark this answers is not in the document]
Siemienik: Well, I think that for version 3.* we should mark this function as `OBSOLETE`/`DEPRECETED` first with message explain when it will be removed (from v4 -may 2020 ) and how to upgrade (link to migration instruction - now we haven't even instruction for that).

I little worry that exceljs users may have headache after upgrade to v4 and finally stay in v3  

I see what awesome work you did and how it's helpful for using xlsx files. I want to push the merge button ASAP. These changes will make a big step to the future 😄 

So, remain, I think we should:
1. Obsolete this function in v3.9.3 (it hasn't published yet :/ )
2. Add any instruction what exactly should do after upgrade to v4.
3. Schedule time when we merge and publish.


orr.. make some adapter for `createInputStream` (if possible), and obsolete (or not) in v4.

README_zh.md
```diff
@@ -1666,6 +1669,83 @@ workbook.commit()
   });
 ```
 
+##### Streaming XLSX Reader
```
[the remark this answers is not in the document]
Alanscut: @Siemienik I'm happy to finish this work.

README.md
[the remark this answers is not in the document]
alubbe: I think that's a great idea - I'll create a PR deprecating it in 3.9.3. How should we mark it? Using `console.log` each time it gets called?

lib/stream/xlsx/workbook-reader.js
alubbe: let's go with 
```js
    this.options = {
      worksheets: 'emit',
      sharedStrings: 'cache',
      hyperlinks: 'ignore',
      styles: 'ignore',
      entries: 'ignore',
      ...options,
    };
```

README.md
alubbe: It's cached into the workbook instance and then, whenever a shared string is found inside a cell, it sets that string:
```js
                        if (sharedStrings) {
                          cell.value = sharedStrings[index];
                        } else {
                          cell.value = {
                            sharedString: index,
                          };
                        }
```

alubbe: I updated the README - do you think it's clearer now?

README_zh.md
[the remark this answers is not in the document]
alubbe: @Alanscut that would be fantastic! I think the README is ready to be translated now

README.md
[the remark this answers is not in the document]
alubbe: Done, using `console.warn`: https://github.com/exceljs/exceljs/pull/1209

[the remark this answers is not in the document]
Siemienik: https://jsdoc.app/tags-deprecated.html

alubbe:
I found and fixed another issue with the zip iteration (see https://github.com/ZJONSSON/node-unzipper/issues/191) and have also deprecated CSV#createInputStream

alubbe:
I've resolved all merge conflicts, this PR is now ready to be merged

spec/integration/workbook-xlsx-reader.spec.js
```diff
@@ -32,13 +32,13 @@ describe('WorkbookReader', () => {
...
             err => {
-              expect(err.message).to.equal('Max row count exceeded');
+              expect(err.message).to.equal('Max row count (10) exceeded');
             }
           );
       });
 
       it('should fail fast on a huge file', function() {
-        this.timeout(20000);
+        this.timeout(5000);
```
alubbe, line 41: This test is currently super flaky on `master`, this PR fixes it - and so confidently, that we can decrease the timeout from 20s to 5s

alubbe:
@guyonroche can we please merge this?

PR #1140, opened 2020-02-21
Optimize SAXStream

alubbe:
This PR brings over improvements and speed-ups from https://github.com/exceljs/exceljs/pull/1135 that work on nodejs v8. It will help make the diff of the other PR smaller


alubbe:
Could we try to review/merge this PR? If anything here is unclear, is there anything I can improve?

Siemienik:
@alubbe I'm going to review when I will have some time

lib/utils/sax-stream.js
```diff
@@ -1,14 +1,36 @@
...
-const saxes = require('saxes');
+const Stream = require('readable-stream');
+const {SaxesParser} = require('saxes');
 
-module.exports = class SAXStream extends Writable {
-  constructor() {
-    super();
-    this.sax = new saxes.SaxesParser();
+module.exports = class SAXStream extends Stream.Transform {
+  constructor(eventTypes) {
```
Siemienik: I see that `eventTypes` everywhere has the same value, as `sax-stream.js` is an internal utility I think you should use default value here.

Siemienik:
image
perfect! 😄

lib/utils/sax-stream.js
alubbe: Good catch - fixed in https://github.com/exceljs/exceljs/pull/1135

PR #1142, opened 2020-02-21
[MAJOR VERSION] Move node v8 support to ES5 imports

alubbe:
This PR drops support for nodejs v8 and bumps the version to 4.0.0
v8 has been unsupported for almost 2 months now and we can use the major version change to start using node v10+ features like async iteration - what do you think?


Siemienik:
hmm, that true

but, a lot of our users still uses node v8 - we should think about how to care about them :)

alubbe:
Okay let's try a totally different approach then :)
I propose to move node v8 support to ES5 imports - users of the old version will get a clear message of what they need to change and everyone on node v10 and newer will get the full power and speed of newer features such as async iteration.

This is strictly speaking a semver-incompatible change, but I think the upgrade path is simple enough - what do you think?

alubbe:
@guyonroche what's your take on supporting outdated versions of node? How often would you like to release a new major version of exceljs?

alubbe:
I've also restored testing v8 on travis, so now the only breaking change is that users of node v8 will get a message to change their `require` statement. Can we merge that and publish it as 4.0.0 @guyonroche ? Then I can finish my work on async iteration and stream cancellation

Siemienik:
We have to publish v3.8.3 from current master first - 3.8.2 has bug inside - #1179 #1180 #1178

Siemienik:
WIP because, 3.8.2 doesn't work now, we have to publish 3.8.3 version before v4.

alubbe:
I've resolved all merge conflicts and have added testing for node 14

Issue #1057, opened 2019-12-19
addConditionalFormatting is not a function error when using Streaming XLSX Writer


werkilla:
addConditionalFormatting function is not available for streams.

code:
```javascript
    const options = {
      filename: './streamed-workbook.xlsx',
      useStyles: true,
      useSharedStrings: true
    };
  
    let workbook = new Excel.stream.xlsx.WorkbookWriter(options);
    let worksheet = workbook.addWorksheet('My Sheet');
    
    worksheet.addRow([1, 2, 3]);
    worksheet.addConditionalFormatting({
      ref: 'A1:E7',
      rules: [
        {
          type: 'expression',
          formulae: ['MOD(ROW()+COLUMN(),2)=0'],
          style: {fill: {type: 'pattern', pattern: 'solid', bgColor: {argb: 'FF00FF00'}}},
        }
      ]
    });
    workbook.commit();```

PR #1143, opened 2020-02-24
fix issue #1057 Fix addConditionalFormatting is not a function error when using Streaming XLSX Writer

Alanscut:
## Change
* fix for issue #1057 
* Support for conditional formatting in xlsx write streams


alubbe:
Looks great but I have no idea about whether the output XML is correct - @guyonroche would you mind taking a look at this one?

Issue #731, opened 2019-01-22
Some media has been lost when read&write file


Siemienik:
I have `jpg` and `png` images putted into xlsx file, all `.png` has been lost after read and write file.

PR #1148, opened 2020-02-27
fix issue #731 image extensions not be case sensitive

Alanscut:
## Changes

* fix for issue #731 
* The cause of this issue #731 is that the suffix of the picture is uppercase letters,which causes the regular match to fail. The drawing1.xml file is not generated when the xlsx file is rewritten, so I personally think that the suffix name should not be case sensitive.


Siemienik:
@Alanscut Thank you!

PR #1159, opened 2020-03-09
Supports setting cell comment properties

Alanscut:
### changes
* The purpose of this PR is to support the setting of cell comment properties. Currently only the protection, margin and object position properties are supported.
* Please let me know if this PR needs anything else.


Siemienik:
@Alanscut, many changes are here, could we make a review session through zoom.us?

Alanscut:
@Siemienik  Sorry for replying so late. I am looking forward to video communication with you, but English is not my native language, video communication will be very difficult for me. I would rather explain my purpose and plan in writing, hope you not mind.

## Purpose:
excel.js currently only supports inserting annotations, but does not allow users to set the corresponding properties of annotations. This PR is to solve this problem.

## Code:
This PR will support setting protection, margin and object position attributes. I also want to support font, alignment, size and color and line attributes, so that users can insert comments flexibly.

* lib / utils / under-dash.js
```deepMerge ()``` is to deeply merge the attributes of the source object to the current object without interfering with each other.

* lib / xlsx / xform / comment / style / vml-position-xform.js
This file is to support location object properties.

* lib / xlsx / xform / comment / style / vml-protection-xform.js
This is to support protection attributes.

* lib / xlsx / xform / comment / vml-textbox-xform.js
This is to support margin attributes.

**This PR is a new feature and will not affect the original function, and all test cases are successful.**

## Demo
```javascript
const Excel = require('../../lib/exceljs.nodejs');
const ws2 = workbook.addWorksheet('sheet1');
ws2.getCell('A1').value = 'Hello';
ws2.getCell('A1').note = 'Hello, ExcelJS!';
ws2.getCell('A5').value = 'world';
ws2.getCell('A5').note = {
  texts: [{
    'font': {
      'size': 12,
      'color': {
        'theme': 0,
      },
      'name': 'Calibri',
      'family': 2,
      'scheme': 'minor',
    },
    'text': 'hello exceljs',
  }],
  margins: {
    insetmode: 'custom',
    inset: [0.55, 0.55, 0.45, 0.45],
  },
  protection: {
    locked: 'True',
    lockText: 'True',
  },
  editAs: 'absolute',
};
const fileName1 = './read/issue_comments_padding.xlsx';
workbook.xlsx.writeFile(fileName1).then(() => {
  console.log('ok');
});
```

## XML

```xml
// vmlDrawing1.vml
<v:shape id="_x0000_s1026" type="#_x0000_t202" style="position:absolute; margin-left:105.3pt;margin-top:10.5pt;width:97.8pt;height:59.1pt;z-index:1;visibility:hidden" fillcolor="infoBackground [80]" strokecolor="none [81]" o:insetmode="custom">
    <v:fill color2="infoBackground [80]" />
    <v:shadow color="none [81]" obscured="t" />
    <v:path o:connecttype="none" />
    <v:textbox style="mso-direction-alt:auto" inset="5.5mm,5.5mm,4.5mm,4.5mm">
      <div style="text-align:left" />
    </v:textbox>
    <x:ClientData ObjectType="Note">
      <x:MoveWithCells />
      <x:SizeWithCells />
      <x:Anchor>1, 6, 3, 14, 3, 2, 7, 16</x:Anchor>
      <x:Locked>True</x:Locked>
      <x:AutoFill>False</x:AutoFill>
      <x:LockText>True</x:LockText>
      <x:Row>4</x:Row>
      <x:Column>0</x:Column>
    </x:ClientData>
  </v:shape>
```

lib/xlsx/xform/comment/style/vml-position-xform.js
```diff
@@ -0,0 +1,39 @@
...
+    } else if (this.tag === 'x:SizeWithCells' && model === type[1]) {
+      xmlStream.leafNode(this.tag);
+    }
+  }
+
+  parseOpen(node) {
+    switch (node.name) {
+      case this.tag:
+        this.model = {};
+        this.model[this.tag] = !!node.name;
```
Siemienik, line 25: is it always true?

spec/integration/pr/test-pr-896.spec.js
```diff
@@ -22,11 +22,11 @@ describe('pr related issues', () => {
         .then(wb2 => {
           const ws2 = wb2.getWorksheet('foo');
           expect(ws2.getCell('A1').value).to.equal(' leading');
-          expect(ws2.getCell('A1').note).to.equal(' leading');
+          expect(ws2.getCell('A1').note.texts[0].text).to.equal(' leading');
```
Siemienik, line 25: could I please you to move your logic from `note` to `notes`  for backward compatibility purposes?
Now, this PR changes the library's API (only for getting a value of course).  `note` will return the first element of notes. 

In other cases, these changes should be published with version 4 and some migration instruction will be added

Alanscut: Thank you @Siemienik I re-modified the logic of `note.js`, and now I just added the corresponding property settings on the original basis, without changing the original way of getting the value. It will always return the value set by the user:).

lib/xlsx/xform/comment/style/vml-position-xform.js
Alanscut: Yes, but it seems a bit ambiguous, I have modified:).

index.d.ts
```diff
@@ -1,3 +1,4 @@
+import { Comment, Protection } from './index.d';
```
Siemienik: what is that?

lib/utils/under-dash.js
```diff
@@ -95,6 +97,46 @@ const _ = {
...
+
+  isArray(val) {
+    return toString.call(val) === '[object Array]';
+  },
+
+  isObject(val) {
+    return toString.call(val) === '[object Object]';
+  },
+
+  deepMerge() {
```
Siemienik, line 113: what is the difference between `Object.assign()`?

lib/xlsx/xform/comment/style/vml-position-xform.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  get tag() {
+    return this._model && this._model.tag;
+  }
+
+  render(xmlStream, model, type) {
+    if (model === type[2]) {
+      xmlStream.leafNode(this.tag);
+    } else if (this.tag === 'x:SizeWithCells' && model === type[1]) {
```
Siemienik, line 16: this if is unnecessary, both variants are the same.

lib/xlsx/xform/comment/vml-notes-xform.js
```diff
@@ -92,6 +100,6 @@ VmlNotesXform.DRAWING_ATTRIBUTES = {
   'xmlns:x': 'urn:schemas-microsoft-com:office:excel',
 };
 
-VmlNotesXform.vmlCommentXform = new VmlNoteXform();
+// VmlNotesXform.vmlCommentXform = new VmlNoteXform();
```
Siemienik: It should be deleted

lib/utils/under-dash.js
Alanscut: The `Object.assign()` method copies all enumerable own properties from one or more source objects to a target object. It returns the target object.The property value of the source object is a reference of an object, so it only points to that reference.

This `deepmerge()` will be deeply copied to the target object, similar to the concept of deep cloning.

lib/xlsx/xform/comment/style/vml-position-xform.js
Alanscut: @Siemienik This is necessary to distinguish the three situations of location attribute.

alubbe:
@Siemienik what do you think?

Issue #204, opened 2016-10-24
Columnwidth of 0 zero when setting numFmt


Sjonnie2nd:
First thanks for the great work.

I'm new to this project so perhaps I miss some experience but..
When setting numFmt of a column: sheet.getColumn(3).numFmt = '0.00', the formatting works as expected but the column gets invisible. (using MS Office Excel 2010). 
Only if also the width is explicitely set: sheet.getColumn(3).width = 10 the column gets visible again.

Is this by design? Wouldn't it be good to warn for this issue in the docs?

LukevdPalen:
@Sjonnie2nd i'm having the same issue. The problem also occurs when applying styles on columns.

Alanscut:
@Siemienik I have researched the code and found that this width attribute is missing from the col tag in **sheet1.xml**. I think we should give a default value when then user does not set the width value. If you agree with this proposal, I would be happy to submit a PR.

### sheet1.xml
```xml
<cols>
    <col min="3" max="3" style="1" customWidth="1" />
</cols>
```

### Source Code
```javascript
  static toModel(columns) {
    // Convert array of Column into compressed list cols
    const cols = [];
    let col = null;
    if (columns) {
      columns.forEach((column, index) => {
        if (column.isDefault) {
          if (col) {
            col = null;
          }
        } else if (!col || !column.equivalentTo(col)) {
          col = {
            min: index + 1,
            max: index + 1,
            width: column.width || 8,   // it's here
            style: column.style,
            isCustomWidth: column.isCustomWidth,
            hidden: column.hidden,
            outlineLevel: column.outlineLevel,
            collapsed: column.collapsed,
          };
          cols.push(col);
        } else {
          col.max = index + 1;
        }
      });
    }
    return cols.length ? cols : undefined;
  }
```

Siemienik:
Is the default value set for worksheet/workbook?

PR #1160, opened 2020-03-10
fix issue #204 sets default column width

Alanscut:
## Changes

* When user-defined column properties, but the column width is not set, we should set a default value for the column box.

* Fix for issue #204 .

### DIFF
--- a/lib/doc/column.js
+++ b/lib/doc/column.js
@@ -5,6 +5,8 @@ const _ = require('../utils/under-dash');
 const Enums = require('./enums');
 const colCache = require('../utils/col-cache');
 
+const DEFAULT_COLUMN_WIDTH = 9;
+
 // Column defines the column properties for 1 column.
 // This includes header rows, widths, key, (style), etc.
 // Worksheet will condense the columns as appropriate during serialization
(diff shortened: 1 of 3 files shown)


lib/doc/column.js
```diff
@@ -5,6 +5,8 @@ const _ = require('../utils/under-dash');
 const Enums = require('./enums');
 const colCache = require('../utils/col-cache');
 
+const DEFAULT_COLUMN_WIDTH = 9;
```
Siemienik: I'm really happy to read your pull requests, good job. I had some issues with width before, so I feel twice satisfaction with your changes 👍 
but I'm not sure, that it's the correct way to set a const value, 
I think that should be property for worksheet and/or workbook, with a possibility to change it by a user.

May I'm wrong, so please clear me this and tell what are you thinking? 😄

Siemienik: Event if I wrote a comment, I think these PR fix some issues anyway, so I approve it to merge 👍

Alanscut: @Siemienik First of all, thank you very much for your approval, this will become my motivation. I really like exceljs, it is very helpful for my work, so I am happy to contribute a little bit of my own power to this software.

* When the user does not set the column width, I will set a default value for the column.

```javascript
// single column setting
 sheet.getColumn(3).width = 10;
```

* Users can globally set the column width of the entire worksheet

```javascript
worksheet.properties.defaultColWidth= 47;
```

alubbe:
@Siemienik should we merge this?

Siemienik:
I confirm that it works ok with [`xlsx-renderer`](https://github.com/siemienik/xlsx-renderer)

PR #1164, opened 2020-03-12
Include cell address for Shared Formula master must exist.. error

noisyscanner:
## Summary
I kept getting the following error while deleting rows from a sheet: `Shared Formula master must exist above and or left of clone`.
I needed to know what cell it was complaining about, so I have changed it to: `Shared Formula master must exist above and or left of clone for cell <cell address>`

## Test plan
It's just a change in text in the error so not sure if it needs a test?

image


alubbe:
Looks great, thanks!

Issue #1177, opened 2020-03-24
[BUG] exceljs throws an error "Invalid range in character set" in IE11 due to the lib "xmlchars"


JamesChenX:
## 🐛 Bug Report

Lib version: 3.8.2

## Steps To Reproduce

1. In IE11
```javascript
require('regenerator/runtime');
const Excel = require('exceljs/dist/es5/exceljs.bare');
```
2. Open the page that uses exceljs in IE11

## The expected behaviour:

exceljs should work fine in IE11

## Possible solution (optional, but very helpful):
The root cause is that exceljs uses saxes that uses xmlchars.
And xmlchars uses regular expressions with the unicode flag ("u") that cannot work in IE (so an error "Invalid range in character set" is thrown)

new RegExp("^["+t.NAME_START_CHAR+"]$","u")

```javascript

```

tonchmx:
I'm trying to use this babel plugin: https://babeljs.io/docs/en/babel-plugin-proposal-unicode-property-regex

Actually, there's the same issue in the xmlchars package: https://github.com/lddubeau/xmlchars/issues/3

tonchmx:
I've rolled back to the 3.8.0 version in my project and I am able to use the exceljs library again

tonchmx:
I think that saxes isn't fully compatible with IE 11? I tested with 3.8.1 too and had the same error.

https://github.com/exceljs/exceljs/pull/1127

alubbe:
Thanks for reporting this. As per https://github.com/exceljs/exceljs/pull/1193, you'll need to polyfill unicode support in IE11, for example like this (before requiring exceljs):

```js
const rewritePattern = require('regexpu-core');
const {generateRegexpuOptions} = require('@babel/helper-create-regexp-features-plugin/lib/util');

const {RegExp} = global;
try {
  new RegExp('a', 'u');
} catch (err) {
  global.RegExp = function(pattern, flags) {
    if (flags && flags.includes('u')) {
      return new RegExp(rewritePattern(pattern, flags, generateRegexpuOptions({flags, pattern})));
    }
    return new RegExp(pattern, flags);
  };
  global.RegExp.prototype = RegExp;
}
```

alubbe:
The above fixed the issue on my windows laptop running IE11

PR #1193, opened 2020-04-01
Ensure all node_modules are compatible with IE11

alubbe:


alubbe:
This PR ensures that 
1. all node modules get transpiled to es5 (needed for fast-csv and saxes)
2. unicode regex patterns are supported (needed for a dependency of saxes)

I've tested this on IE11 and can confirm that it fixes https://github.com/exceljs/exceljs/issues/1177

PR #1195, opened 2020-04-01
Allow user specified iteration count

mah110020:
## Summary

The iteration count controls how many times the hash is applied. It is a mechanism for fine-tuning according to how much computation is desired. This is especially useful to future-proof against increasingly powerful computers (you can simply increase the iteration count). It is therefore natural to expose this as a setting to the developer.

In my use case, I use this library for generating web application downloads, and protecting each worksheet in the workbook caused a download delay of several seconds. I would like to be able to lower the iteration count.

## Test plan

```
const time1 = new Date();
await sheet.protect("password");
const time2 = new Date();
await sheet.protect("password", { spinCount: 1000 });
const time3 = new Date();

console.log(`Delta 1: ${time2 - time1}`); // < 1 second
console.log(`Delta 2: ${time3 - time2}`); // ~0 seconds
```

I have created a test script which generates excel documents for various `spinCount` values, and have confirmed that excel opens them all, that the sheet is protected, and that I am able to unprotect with the correct password. And while I can't time how long it took excel to unlock (or fail to unlock), the time seemed to be roughly equal to the protection time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1860,7 +1860,9 @@ to modify individual cell protection.
 
 **Note:** While the protect() function returns a Promise indicating
 that it is async, the current implementation runs on the main
-thread and will use approx 600ms on an average CPU.
+thread and will use approx 600ms on an average CPU. This can be adjusted
+by setting the spinCount, which can be used to make the process either
+faster or more resilient.
 
 ### Sheet Protection Options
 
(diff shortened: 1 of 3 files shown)


mah110020:
Does the `spinCount` test need to handle only those kinds of values, or should `spinCount` be able to handle other kinds of values too? For example: Infinity, NaN, floating points, etc.?

Siemienik:
It always is better when testing more than less:)

mah110020:
I'm not completely sure how to integrate the requested tests, but I have made the requested changes.

PR #1203, opened 2020-04-06
Add null type to TableStyleProperties.theme interface

sijmenvos:
## Summary

According to the [docs](https://github.com/exceljs/exceljs#table-style-themes):

> For no theme, use the value null.

But having `style: { theme: null }` in `addTable` gives:
> Type '{ theme: null; }' is not assignable to type 'TableStyleProperties'.

## Test plan

Only a minor change, Typescript compiles without issues and result in Excel is working as well.


fer22f:
Why was this closed? This needs to be merged.

PR #834, opened 2019-05-25
add cn doc

loverto:

PR #1208, opened 2020-04-08
Re-translation of simplified Chinese documents (zh-cn)

wang1212:
I re-translated the simplified Chinese document (zh-cn) by myself. The following are the reasons why I did this:

- When I used the `xlsx` package in my project, I encountered that exporting xlsx files could not add styles (it is said to be only available in the pro version). I found` exceljs` as an alternative and it was very easy to use.
- I found a Chinese document (#834 ) while viewing the document, but after reading it, I found it very disappointing, and the traces of machine translation are also obvious.

Therefore, I plan to re-translate to avoid misleading others. I am not a professional translator, but the documents will be corrected manually sentence by sentence. Due to the limitations of my personal level, errors inevitably occur. However, I can guarantee that most of them are correct.

---

The body part of the document has been completely translated, and the "Release History" part has not been completely translated yet. I will take the time to improve later.

### DIFF
--- a/README_zh.md
+++ b/README_zh.md
@@ -4,9 +4,9 @@
 [![Code Quality: Javascript](https://img.shields.io/lgtm/grade/javascript/g/exceljs/exceljs.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/exceljs/exceljs/context:javascript)
 [![Total Alerts](https://img.shields.io/lgtm/alerts/g/exceljs/exceljs.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/exceljs/exceljs/alerts)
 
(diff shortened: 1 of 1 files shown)


README_zh.md
```diff
@@ -4,162 +4,245 @@
...
+# 接口
+
+## 创建工作簿
 
 ```javascript
 ```
 
-## <a id="set-workbook-properties">设置工作簿属性</a>
+## 设置工作簿属性
```
Alanscut, line 245: @wang1212 Removing the `a` tag will cause the anchor to fail. Is your local verification successful?

wang1212: Yes, I don’t need the `a` tag to locate the anchor point locally. Moreover, you can see that it is normal in my git repository [wang1212/exceljs](https://github.com/wang1212/exceljs/blob/master/README_zh.md)

PR #1209, opened 2020-04-08
Deprecate createInputStream

alubbe:
As discussed in https://github.com/exceljs/exceljs/pull/1135, let's deprecate `createInputStream` in favour of the `read` methods in 4.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -1932,21 +1932,20 @@ faster or more resilient.
 ```javascript
 // read from a file
 var workbook = new Excel.Workbook();
-workbook.xlsx.readFile(filename)
-  .then(function() {
-    // use workbook
-  });
+await workbook.xlsx.readFile(filename);
+// ... use workbook
 
-// pipe from stream
+
+// read from a stream
 var workbook = new Excel.Workbook();
-stream.pipe(workbook.xlsx.createInputStream());
+await workbook.xlsx.read(stream);
+// ... use workbook
+
 
 // load from buffer
 var workbook = new Excel.Workbook();
-workbook.xlsx.load(data)
-  .then(function() {
-    // use workbook
-  });
+await workbook.xlsx.load(data);
+// ... use workbook
 ```
 
 #### Writing XLSX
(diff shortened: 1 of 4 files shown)


guyonroche:
@alubbe @Siemienik Awesome work you guys have put in here! It's time to publish again - this PR making for major version. Is there anything else you think should go in first?
PS. Hope you are all keeping well during this pandemic.

alubbe:
I think we have everything on `master` to deploy one last bugfix release on v3, and then we should merge https://github.com/exceljs/exceljs/pull/1142 and https://github.com/exceljs/exceljs/pull/1135 (in that order) to push out v4

guyonroche:
@alubbe @Siemienik just a couple of questions...
This PR here is a breaking change (dropping an API function) so will mean a major version (i.e. v4), Given that #1142 and #1135 are the way forward, do we want to merge them before v4 or wait for v5?

I noticed somewhere along the line about 22 days ago that the .travis.yml file got dropped - was this intentional?

Siemienik:
We planned to publish it in v3, 

>I noticed somewhere along the line about 22 days ago that the .travis.yml file got dropped - was this intentional?
I added github workflow for checking if index.d.ts is compiled correctly:
https://github.com/exceljs/exceljs/pull/1182

But it has side-effect - it's turn off travis checks 

So @alubbe  made this PR with migration from travis to github workflows: https://github.com/exceljs/exceljs/pull/1190

Alanscut:
@alubbe @Siemienik I have to say your work is great, but I found that there are still some problems with index.d.ts.
1. The file is too large, which is not conducive to maintenance and use, and should be split according to the module
2. The file is missing a type declaration, for example: `WorkbookReader`,` WorksheetWriter`, `WorksheetReader`, etc.
3. Private methods should not be exposed, for example: `addWorkbookRels`,` addWorkbook` and `addSharedStrings` in` WorkbookWriter`

This is the pain point of `exceljs`, and it has caused a lot of trouble to developers. Personally, before releasing the main version, we should review` index.d.ts` and solve the above problems. Waht do you think?

alubbe:
I'm not a TS user myself, so I have no idea about any of this - but personally, I think we've never guaranteed that `index.d.ts` works perfectly, so I think doing those fixes (which sound great) can be done after publishing 4.0.0. I would just really for us to publish 4.0.0 soon, since the open PRs are pretty big

Siemienik:
Well, I will be a big fan of migrating to TS (with --allowJs) and step by step move our logic form js files to ts 😄 

I feel a big lack of types, now I have to investigate the whole project to do something because IDE doesn't provide intellisense😭. 

I wonder about create milestones, fo version 4.1, 4.2 and 5. It should help us to keep one vision of further developing this library.

What is your opinion?

alubbe:
I'd also like us to move to Typescript, but 
a) I don't think we should delay moving to 4.0.0
b) let's create a new issue to discuss typescript/index.d.ts instead of discussing here on this closed issue ;)

Issue #1198, opened 2020-04-02
[BUG] Loading OpenPyXL workbooks


mannershark:
## 🐛 Bug Report

I've created excel files using OpenPyXL and now want to modify these using ExcelJS. However, there are no sheets on the Workbook.worksheets array.
There was a single worksheet located at Workbook._worksheets on the `undefined` property, but the others were missing.

I did some debugging, and found that the issue occurs at `xlsx.js::353`, `this.workbook.model = model`. The model properly loads and contains all the worksheets, but the setter doesn't handle them correctly.

In `workbook.js::200`, `const {id, name, state} = worksheetModel;` has `id` undefined, so at some point the worksheet id is missing or not set properly. This may be caused by OpenPyXL not saving the worksheet properly?

Lib version: 3.8.2
Openpyxl: 3.0.3

## Steps To Reproduce

Python
```python
    wb = Workbook()
    example_row = ['1', '2', '3']
    ka_list = wb.create_sheet('KA lijst')
    ka_list.append(example_row)
    wb.save(f"output/example.xlsx")
```

JS
```javascript
    workbook = new Workbook();
    await workbook.xlsx.readFile(__dirname + 'output/example.xlsx');
    const kaWorksheet = workbook.getWorksheet('KA lijst');
    // kaWorksheet is undefined
```
File with the incorrect behaviour:
[lasschema.xlsx](https://github.com/exceljs/exceljs/files/4421247/lasschema.xlsx)

## The expected behaviour:

I would expect to be able to retrieve the specific sheet using `getWorksheet('KA')` and be able to read its rows.

## Potential fix:

In `workbook-xform.js::160` we find the worksheet using `rel.Target`. ( `worksheet = model.worksheetHash[`xl/${rel.Target}`]`) 
For the given file, `rel.Target` is `"/xl/worksheets/sheet1.xml"`, including a '/xl/' already. The lookup with `xl/xl/worksheets/sheet1.xml` therefore won't find it.

For example
```
  if (rel.Target.startsWith('/xl/')) {
    rel.Target = rel.Target.substr(4);
  }
```

We could slice the "/xl"section off rel.Target when it's present to fix it.

Alanscut:
Hi @mannershark 
See #1185，Have the same error message.

mannershark:
Confirmed fixed! Thanks @Alanscut for the PR and @alubbe for reviewing

PR #1220, opened 2020-04-17
Fix issue #1198 Absolute path and relative path need to be compatible

Alanscut:
## Changes 
* The target attribute of workbook.xml.rels has absolute path and relative path, this path problem needs to be compatible.
* Fix fro issue #1198 .


lib/xlsx/xform/book/workbook-xform.js
```diff
@@ -157,7 +157,7 @@ class WorkbookXform extends BaseXform {
       if (!rel) {
         return;
       }
-      worksheet = model.worksheetHash[`xl/${rel.Target}`];
+      worksheet = model.worksheetHash[`xl/${rel.Target.replace(/^(\s|\/xl\/)+/, '')}`];
```
alubbe, line 160: could you please add a comment here as to what strings would be transformed here? I.e., anything with a preceding `/xl/`?

Alanscut: @alubbe yes, if `rel.Target` start with `[space]/xl/` or `/xl/`  , then it will be replaced with `''` and spliced behind  `xl/`,  otherwise it will be spliced directly behind `xl/`.  i.g.
```
'/for/bar' ---> 'xl/for/bar'
'   /xl/for/bar' ---> 'xl/for/bar'
'/xl/for/bar' ---> 'xl/for/bar'
```

spec/integration/pr/test-pr-1216.spec.js
```diff
@@ -0,0 +1,10 @@
+const ExcelJS = verquire('exceljs');
+
+describe('github issues', () => {
+  it('pull request 1216 - The worksheet should not be undefined', async () => {
```
alubbe, line 4: shouldn't this be PR #1220 ? PR looks good otherwise

Alanscut: @alubbe I update the test case. thanks :).

Issue #1131, opened 2020-02-14
headerFooter property doesn't work with WorksheetWriter


asivov:
Hi dear team,
I'm using the 3.8.0 version of exceljs and when I'm trying to set the headerFooter property for WorksheetWriter, it doesn't work. So it works perfectly when we are using Worksheet, but doesn't work for streams. I've investigated the code and found that there is no HeaderFooterXform using in WorksheetWriter.
I already prepared the fix for this issue. Could you be so kind to give me permissions to prepare the pull request with fix.

Siemienik:
Hi @asivov , it's nice to read that you prepare fixation:smile: 
if could you, please fork this repo and later make there fix for this issue.
The last step will be to create a pull request, which we will merge after code review was done.

PR #1232, opened 2020-04-24
Fix issue #1131 headerFooter property doesn't work with WorksheetWriter!

Alanscut:
# Changes
* Fix headerFooter property not applicable to WorksheetWriter.
* Fix issue #1131 .


Siemienik:
Waiting for checks results and may was merged

Issue #1236, opened 2020-04-27
[BUG] Huge size increase in 3.8.2 -> 3.9.0


mydea:
## 🐛 Bug Report

I wasn't quite sure if this would be a bug or a feature improvement, but I noticed that the jump from 3.8.2 to 3.9.0 lead to an increase of 156 kb, or 36kb gzipped, of the size of the library. 

You can also see this here:
https://bundlephobia.com/result?p=exceljs@3.9.0

From scanning the commits, it is not clear to me what feature lead to this quite massive increase - I am also not sure if that was "on purpose" or just happened somehow.

Lib version: 3.9.0

## Steps To Reproduce

See: https://bundlephobia.com/result?p=exceljs@3.9.0
Which was also confirmed by our CI, which shows the bundle size increasing massively.

## The expected behaviour:

A minor version bump that is not introducing massive new features shouldn't increase the bundle size by over 150 kb.

## Possible solution (optional, but very helpful):

* What exactly lead to this bundle size increase?

alubbe:
hi @mydea thanks for the report! I'm not sure this tool is a good measure, it seems to `npm install` and measure the folder size. We bundle a lot of stuff, incl. the actual source code, a few ES5 transpiled versions and docs. Feel free to investigate if the (es5) browser bundle has seen any size increases, then we can investigate that further.

mydea:
We get the same size increase in our application build output!

alubbe:
Looked into it, it seems to be caused by babel: https://github.com/babel/babel/blob/master/CHANGELOG.md#v795-2020-04-07
The two changes to `createSuper` print a lot of new code that is replicated for each file in our code base and every dependency. It might be that there is a smarter way to not replicate these babel helpers as much - if you have a suggestion to improve our build setup, I'd love to try that out: https://github.com/exceljs/exceljs/blob/master/gruntfile.js

mydea:
Hmm, I have tried downgrading @babel/core all the way down to 7.5 and always get the same output file size, so I don't think it is (only) that which causes the size increase.

I have checked it out locally, and the gzipped size of the excel.min.js increased here:

Baseline 2.8.2: 231kb

* https://github.com/exceljs/exceljs/pull/1140 231kb -> 241kb
* https://github.com/exceljs/exceljs/pull/1193 241kb -> 264kb

[...]

PR #1243, opened 2020-04-30
Compare asset sizes on pull requests

mydea:
## Summary

This PR adds a Github Action that runs on PRs that compares the build output asset sizes with master. This way, it is much easier to notice if a PR adds considerable size to the library.

This should help with issues like https://github.com/exceljs/exceljs/issues/1236 in the future.

It looks like this in action:
https://github.com/mydea/exceljs/pull/2

### DIFF
--- a/.github/workflows/asset-size.yml
+++ b/.github/workflows/asset-size.yml
@@ -1,0 +1,18 @@
+name: Asset size
+
+on: [pull_request]
+
+jobs:
+  compare:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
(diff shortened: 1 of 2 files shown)


mydea:
Note: The asset size action itself will fail on this PR as the `npm install-build` command is not available on the master branch yet - so that is to be expected.

The other test failures are not clear to me...

alubbe:

mydea:
I have rebased this on top of master now!

alubbe:
Why is this failing with `npm ERR! missing script: install-build`? The script seems to be defined in the package.json

wereHamster:
@alubbe but not in the master branch. This action builds the repo twice, once master and once the current version, so it's able to compare the two. If the build script is not available in master then it won't work.

alubbe:
That makes sense - let's give it a try :)

PR #1259, opened 2020-05-08
Fix hidden

ayatkyo:
## Summary
Add more check for `hidden` attribute in Row or Column. Previously `hidden: 'false'` considered as `true` that make row and column in output file to be hidden.

## Test plan
Test case added to `spec\integration\worksheet.spec.js`


alubbe:

Siemienik:
@ayatkyo  thanks

Issue #1261, opened 2020-05-08
[BUG] WorkbookWriter sheet.protect() function doesn't exist


rheidari:
## 🐛 Bug Report

Unable to add cell level protection on worksheet when using the streaming WorkbookWriter due to sheet.protect() not existing in `worksheet-writer.js`. 

Lib version: 3.9.0

## Steps To Reproduce

```javascript
const workbook = new Excel.stream.xlsx.WorkbookWriter({
     filename: './test.xlsx',
});

const sheet = workbook.addWorksheet('data');
const row = sheet.addRow(['readonly cell']);
row.getCell(1).protection = { locked: true };

expect(sheet.protect).to.exist();
await sheet.protect('password');

await workbook.commit();
```

## The expected behavior:

Should create a file that can be opened and 'readonly cell' should have protection.

## Possible solution (optional, but very helpful):

Add protect method to `worksheet-writer.js`

raeesaa:
Any updates on this issue?

raeesaa:
@rheidari / @Alanscut Any ETA for merging PR for this? Actually we wanted to use this library but this is the only thing that is blocking us.

Alanscut:
I have approved the fix PR, and wait for other members to review it.

raeesaa:
Hi @Alanscut, sorry to bug again about this. Just wanted to know if we have ETA for this as we have first production release next month. And it is important for us to lock excel. If this will take time, we will start working on alternatives. Thank you for the support and great library :)

PR #1262, opened 2020-05-08
Fix issue #1261 WorkbookWriter sheet.protect() function doesn't exist

rheidari:
## Changes

- Added `protect` method to `worksheet-writer.js` (copied from `worksheet.js`)
- Fix for #1261


lib/stream/xlsx/worksheet-writer.js
```diff
@@ -450,6 +451,39 @@ class WorksheetWriter {
     return this._background && this._background.imageId;
   }
 
+  // =========================================================================
+  // Worksheet Protection
+  protect(password, options) {
+    // TODO: make this function truly async
+    // perhaps marshal to worker thread or something
+    return new Promise(resolve => {
```
Alanscut, line 459: Whether this attribute should be declared in `constructor ()`.
```javascript
    // worksheet protection
    this.sheetProtection = null;
```

rheidari: This has been added to the constructor

alubbe:

alubbe:
The implementation looks good, but I think we need a better test here. Ideally, write a file with a protection and then show that modifying it produces an error

rheidari:
> @rheidari, have you possibility to made requested changes?

Apologies for the delay, I've updated the test!

raeesaa:
Any updates on this issue?

Alanscut (approved):

lib/stream/xlsx/worksheet-writer.js
VikasAnand354: Does this really work ? I set sheet protection to true it doesn't lock a particular cell

PR #1291, opened 2020-05-23
Added insert rows functionality

Subhajitdas298:
## Summary

Added convenience method to insert any rows in between existing rows.

## Test plan

Added 3 new test cases for insert row.


Subhajitdas298:
Resolved conflict with latest master

gruntfile.js
```diff
@@ -11,6 +11,7 @@ module.exports = function(grunt) {
     babel: {
       options: {
         sourceMap: true,
+        compact: true,
```
alubbe, line 14: I'm not 100% sure, but I think we don't want this setting here. I believe the idea is that the babel output is nice and readable, and only the minified build is compact

README.md
```diff
@@ -767,6 +767,27 @@ var rows = [
 ];
 worksheet.addRows(rows);
 
+// Insert a couple of Rows by key-value, shifting down rows every time
```
alubbe, line 770: could you please add the same thing to the chinese readme? No need to translate it for now

alubbe:
We just did a group review - once those two points are addressed, this is good to go 👍

gruntfile.js
Subhajitdas298: > I'm not 100% sure, but I think we don't want this setting here. I believe the idea is that the babel output is nice and readable, and only the minified build is compact. Let's take it out of this PR - if you think we should change this, let's do that in a separate PR

In that case make it false? Because, by default babel will attempt to make it compact (upto 500KB). And it will throw that warning, as files are above that size. It would be great to have a defined value, rather than uncertain results.

Subhajitdas298: Changed compact to false

Subhajitdas298: Please review. Thanks

README.md
Subhajitdas298: Done

Subhajitdas298:
Added option to inherit style for addRow/s and insertRow/s without affecting existing implementations. Added test cases for same.
Please review.

Subhajitdas298:
@alubbe Please review the latest changes. Thanks

Subhajitdas298:
@alubbe Please review my recent changes. Thank you

lib/doc/worksheet.js
```diff
@@ -334,18 +334,57 @@ class Worksheet {
...
 
+  insertRow(pos, value, styleOption = 'n') {
+    this.spliceRows(pos, 0, value);
+    this._setStyleOption(pos, styleOption);
+    return this.getRow(pos);
+  }
+
+  insertRows(pos, values, styleOption = 'n') {
+    this.spliceRows(pos, 0, ...values);
+    if (styleOption !== 'n') {
```
alubbe, line 359: @Siemienik I remember we had a discussion around cloning/copying style objects - could you review this part of the new code?

alubbe:
The new changes look great - just left a comment regarding the copying of styles, let's see what @Siemienik has to say

lib/doc/worksheet.js
Siemienik: I think that the best idea to maintain style's is change logic of setters to use Object.freez and copy values there. ( When any setter style is being used then all style object should be copy and freez)

I didn't know yet how this solution impact to this pr.

Subhajitdas298: Style setting logic is based on already existing logic in splice method. 

`rDst.style = rSrc.style;
    // eslint-disable-next-line no-loop-func
    rSrc.eachCell({includeEmpty: true}, (cell, colNumber) => {
      rDst.getCell(colNumber).style = cell.style;
    });
    rDst.height = rSrc.height;`

Are you suggesting something like, this:
`rDst.style = Object.freeze({...rSrc.style});
    // eslint-disable-next-line no-loop-func
    rSrc.eachCell({includeEmpty: true}, (cell, colNumber) => {
      rDst.getCell(colNumber).style = Object.freeze({...cell.style});
    });
    rDst.height = rSrc.height;`

Subhajitdas298: @Siemienik are you saying, to do it like this:
rDst.getCell(colNumber).style = Object.freeze({...cell.style}); });

Subhajitdas298: @Siemienik the existing property specific setter actual logic is in "_applyStyle" in row.js and again in cell's setter. This is not using Object.freeze yet. This may be added in another PR.
This pr directly copies style internally, just like spilceRows. 
I have update the copy logic with Object.freeze as you told.
@Siemienik @alubbe Please review. Thank you.

PR #1297, opened 2020-05-25
Do not include IE11 fix by default, let the users import it when needed

alubbe:
It adds 20KB to the build, which not everyone needs
See https://github.com/exceljs/exceljs/issues/1236


lib/exceljs.browser.js
```diff
@@ -1,26 +1,9 @@
 /* eslint-disable import/no-extraneous-dependencies,node/no-unpublished-require */
 require('core-js/modules/es.promise');
-require('core-js/modules/es.string.includes');
```
Alanscut, line 3: @alubbe we should update READMD.md, https://github.com/exceljs/exceljs#es5-imports--.

alubbe: Good catch - added

PR #1215, opened 2020-04-13
docs: add links to top with jump2header

strdr4605:
## Summary

Links back to Table of contents helps to navigate in the docs.

See [demo](https://github.com/strdr4605/exceljs/blob/jump2header/README.md#importing).

PR #1301, opened 2020-05-26
Revert "add spaces before jump to header links"

strdr4605:
#1215

This reverts commit f65592e6dd5b1a1a1ef61c0c6cba29111a35b40e.

[Demo](https://github.com/strdr4605/exceljs/blob/jump2header/README.md#importing)

Also, in another branch, there is a version with different emoji

[Demo2](https://github.com/strdr4605/exceljs/blob/jump2header2/README.md#importing)

So if you like another option I can update this PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,13 +190,13 @@ To be clear, all contributions added to this library will be included in the lib
   <li><a href="#release-history">Release History</a></li>
 </ul>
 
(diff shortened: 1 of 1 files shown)


alubbe:
Both icons are quite bulky, but I don't see any better option - what do you think @Siemienik @Alanscut ?

PR #1317, opened 2020-06-05
Remove const enum and add ErrorValue in index.d.ts

aplum:
##  Summary

Using `const enum` in .d.ts files breaks compilation with `--isolatedModules`. Instead, normal `enum`s have been used where a corresponding object exists at runtime. `PaperSize` does not exist at runtime, and so has been left as `const enum` for backwards compatibility.

This also adds a declaration for `ErrorValue` since it exists at runtime but had no TS definition. This does result in some duplication with `CellErrorValue`, but changing `CellErrorValue` to use `ErrorValue` would be a breaking change.

A comparable fix and explanation for the problem with using `const enum` can be found at https://github.com/chalk/chalk/pull/296, and some further explanation at https://ncjamieson.com/dont-export-const-enums/ and [this comment](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/23002#issuecomment-359051201).

## Test plan

I created a small project to test this while making changes. Example file:

```typescript
import { ValueType, CellModel } from 'exceljs';

let vt: ValueType = ValueType.Date;
const vtString = ValueType.String;

export const cm1: Partial<CellModel> = { type: vt };
export const cm2: Partial<CellModel> = { type: vtString };
export const cm3: Partial<CellModel> = { type: ValueType.Boolean };
```

Before this PR:
- No errors with `isolatedModules: false`
- Errors with `isolatedModules: true`

With this PR:
- No errors with `isolatedModules: false`
- No errors with `isolatedModules: true`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,6 @@
 declare interface Buffer extends ArrayBuffer { }
 
-export const enum RelationshipType {
+export declare enum RelationshipType {
 	None = 0,
 	OfficeDocument = 1,
 	Worksheet = 2,
@@ -11,7 +11,7 @@ export const enum RelationshipType {
 	Hyperlink = 7
 }
 
-export const enum DocumentType {
+export declare enum DocumentType {
 	Xlsx = 1
 }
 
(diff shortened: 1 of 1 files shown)


alubbe:
@Siemienik do you think we could make the above typescript code into a test file and add it to our test suite?

Siemienik:
@aplum thanks 😄

PR #1328, opened 2020-06-15
bug fix can not read property date1904 of undefined

yganatra1:
## Summary

**Issue**: 
ExcelJs.stream.xlsx.WorkbookReader reading with styles: cache option was giving 
**Cannot read property 'date1904' of undefined**

**Reproduce with this code**
```
const workbookReader = new excel.stream.xlsx.WorkbookReader(fileReadStream, { worksheets: 'emit', styles: 'cache', sharedStrings: 'cache', hyperlinks: 'ignore', entries: 'ignore' });
    workbookReader.read();
    workbookReader.on('worksheet', worksheet => { 
         worksheet.on('row', (row) => {...}
     }
```

Above code will throw an error when `DD/MM/YYYY` format date is given in the excel
e.g: 20/06/2020

I found that in `lib/stream/xlsx/worksheet-reader.js` we are using it but not checking for the undefined model case.


Alanscut (approved):

Siemienik:
LGFM, but it's required to cover each detected (and fixed) bug by test. Could I ask you to write at least one?

Siemienik (changes requested):
Looks good for me, thanks

I have only one request - it's required to cover each detected (and fixed) bug by a test. Could I ask you to write at least one?

mouse9:
Yeah sure.
I understand I was just waiting for your reply first.

I will add test cases now

mouse9:
@Siemienik I have added the integration test case for the same :)

Siemienik (approved):

xcybplx:
For anyone landing here from a search: the cause was in `lib/utils/iterate-stream.js`, which paused the stream around each `yield`. unzipper's entry stream keeps working while paused and emits `end` once its input is exhausted, so entries not yet delivered were dropped silently. The reader hit it through its own workaround — buffering an early worksheet to a temp file — and that write was the delay that lost the rest, including the sharedStrings it was waiting for. Timing-dependent, hence green on Linux and red on macOS/Windows.

gameroman:
I'd recommend trying out one of theseforks

https://github.com/excel-js/exceljs

https://github.com/cjnoname/excelts

https://github.com/protobi/exceljs

You can also see a [comparison on npmx](https://npmx.dev/compare?packages=exceljs,@cj-tech-master/excelts,@excel.js/exceljs,@protobi/exceljs)

Issue #1337, opened 2020-06-17
[BUG] Async Iterators definition for typescript


JuhBoy:
## 🐛 Bug Report

When trying to use `WorkbookReader` as an async iterator with a `for await of` the TSC isn't recognizing the Symbol.asyncIterator property as defined.

Lib version: ^4.0.1

## Steps To Reproduce

```typescript
const workbookReader: IWorkbookReader = new exceljs.stream.xlsx.WorkbookReader(readStream, {});

 for await (const workbookSheet of workbookReader) {}
```
Exception: The `WorkbookReader` type must have a `[Symbol.asyncIterator]()` method which returns an asynchronous iterator.

## The expected behaviour:

No errors should be triggered when using async iterators with an async loop.

## Possible solution (optional, but very helpful):

Change the definition inside the index.d.ts line 1960 && 1974 like so:

```typescript
[Symbol.asyncIterator](): AsyncGenerator<WorksheetReader>;
```
1. parentheses were missing.
2. Due to the implementation in JS code, it would be better to use AsyncGenerator instead of AsyncIterator Generics interfaces.

I could push a pull request if you wan't (and if i'm not wrong about all of this).
Thanks & have a good day.

PR #1338, opened 2020-06-17
Fix: Async iterators defition

JuhBoy:
Hey,

#1337 everything is explained inside this issue. 

Trying to fix/help with this definition file.


Siemienik:
good catch

prandell:
Is this fix in a release yet? It appears version 4.3.0 still has this issue

PR #1346, opened 2020-06-19
rename slave to aligned

stephanos:
## Summary

Slave is an antiquated word and can easily be replaced with something less insensitive. I took inspiration from the [ag-grid project and their improved naming scheme](https://github.com/ag-grid/ag-grid/commit/432bf120cac8b2d050b89f940b31c5bff95abb93).

## Test plan

I can the tests and everything seems to be working just fine. Since I only changed a comment in the actual code and all other changes relate to docs/tests, I feel very confident that I didn't break anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2569,15 +2569,15 @@ Shared formulae enhance the compression of the xlsx document by decreasing the r
 of text within the worksheet xml.
 The top-left cell in a range is the designated master and will hold the
 formula that all the other cells in the range will derive from.
-The other 'slave' cells can then refer to this master cell instead of redefining the
+The other 'aligned' cells can then refer to this master cell instead of redefining the
 whole formula again.
-Note that the master formula will be translated to the slave cells in the usual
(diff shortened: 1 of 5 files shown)


papb:
I support this. Perhaps `child` is also an option (I don't know the context)

Siemienik:
Community lets vote:

👍 for approve
👎 for close

The survey last until 31 October

alubbe:
I'm in favour of renaming this, but I think we should rename `master` as well, at the same time. So let's try to figure out a good new name for this pair.
Old: `master/slave`
New: `primary/replica`... other suggestions?

papb:
* `main cell / sibling cell`
* `reference cell / associated cell`

Siemienik:
* `parent / child`
* as @papb mention : `main / sibling` sounds good

alubbe:
My vote is for `main / sibling` as well

bk521234 (approved):

Issue #1357, opened 2020-06-28
[BUG] 4.1.0 causes TypeScript compilation errors - addRows styleOption should be optional?


cduff:
## 🐛 Bug Report

Upgrading to 4.1.0 results in TypeScript compilation errors for previously working code like:

`sheet.addRows(rows);`

It looks like the problem was introduced in [Added insert rows functionality with new style inherit options](https://github.com/exceljs/exceljs/pull/1324). The new TypeScript definition is:

`addRows(rows: any[], styleOption: string): void;`

Shouldn't styleOption be optional with a default value?

Subhajitdas298:
Sorry for the mishap. And thanks for quick response @sdg9670.
For the code, @cduff, If you are in hurry for release, you can use 'n' in styleOptions parameter.

alubbe:
Closing this, as we just need to wait for the release: #1362

PR #1358, opened 2020-06-29
Fix styleOption error in index.ts

sdg9670:
## Summary

#1357 

Fix styleOption error in index.ts

As typescript updates, errors occur where existing styleOption is omitted.

image


Subhajitdas298:
Thanks for the fix. Sorry, I missed the optional identifier.

Subhajitdas298:
@alubbe Please merge this, as it fixes missing optional operator for type script users.

Issue #1254, opened 2020-05-06
[BUG] getSheetValues() typescript definition is incorrect


danielnitsche:
## 🐛 Bug Report

Lib version: 3.9.0

## Steps To Reproduce

`getSheetValues()` is [defined as](https://github.com/exceljs/exceljs/blob/master/index.d.ts#L1197):
```
	/**
	 * return all rows as sparse array
	 */
	getSheetValues(): Row[];
```

So one would expect a `Row` type to be returned.

## The expected behaviour:

However reviewing the [code for this function](https://github.com/exceljs/exceljs/blob/master/lib/doc/worksheet.js#L444):
```
  // return all rows as sparse array
  getSheetValues() {
    const rows = [];
    this._rows.forEach(row => {
      if (row) {
        rows[row.number] = row.values;
      }
    });
    return rows;
  }
```

A plain array is being returned instead.

## Possible solution (optional, but very helpful):

Update type definition to

```typescript
	/**
	 * return all rows as sparse array
	 */
	getSheetValues(): Cell[][];

```

Alanscut:
The array may be any type of value, string, object, etc.
```javascript
/**
* return all rows as sparse array
*/
getSheetValues(): any[];
```

Siemienik:
@Alanscut we shout avoid using of `any`. It means that we haven't idea what happens.


As we can find here:
https://github.com/exceljs/exceljs/blob/2d8215477be382fc95d475025c0c66c95c44f30a/index.d.ts#L536-L540

the following code should works great:

```ts
type RowValues=CellValue[] | { [key: string]: CellValue }; 
/*...*/
getSheetValues(): RowValues[];
```


P.S. I wrote it in comment without test, please apologize if I made a mistake.

PR #1360, opened 2020-06-30
[bugfix]Fix Issue #1254 and update index.d.ts

Alanscut:
## Changes
* Fix for issue #1254 .


Siemienik:
Is it also possible to return always any array? `[]` instead of null or undefined

PR #1377, opened 2020-07-07
Merge cells after row insert

curtcommander:
## Summary

Merges aren't preserved when cells containing merges get overwritten after row inserts. Instead, the value of a merged cell gets applied to all cells in the merge separately.

| A1 and B1 | C1 |
|---|---|
| merged cell text || 

becomes

| A1 | B1 | C1 |
|---|---|---|
| merged cell text | merged cell text ||

This is because the values for the previously merged cells are pulled from the source row (`rSrc.values` in `spliceRows`). The source row's values repeat the merged value for each cell in the merge. For example, the source row values for the row in the first table would be `["merged cell text", "merged cell text"]`. These values are what gets applied to the cells in the destination row (`rDst`), and the merge is lost.

The extra lines I wrote in `spliceRows` to fix this reapply merges to cells while accounting for the offset caused by a row insert.

## Test plan
I added a test to spec/unit/doc/worksheet.merge.spec for this issue and verified that it passes with the changes I made.

### DIFF
--- a/lib/doc/worksheet.js
+++ b/lib/doc/worksheet.js
@@ -399,7 +399,6 @@ class Worksheet {
     // now copy styles...
     for (let i = 0; i < count; i++) {
       const rDst = this._rows[rowNum + i];
-      rDst.values = rSrc.values;
       rDst.style = rSrc.style;
       rDst.height = rSrc.height;
       // eslint-disable-next-line no-loop-func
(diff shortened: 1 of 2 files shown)


Siemienik (approved):
Nice done

calumpeak:
@curtcommander @Siemienik does this also cover a case where merged cells below rows which have been inserted also exhibit the same behaviour (value of merged cell gets applied to all cells)?

curtcommander:
@calumpeak Which version of exceljs are you using? The current latest version (4.2.0) preserves merges below rows that have been inserted, whereas 4.1.0 does not. I just tested this.

Issue #791, opened 2019-04-24
cell.fill issue


ns20011997:

Siemienik:
@ns20011997 
which version have you used?

ns20011997:
@Siemienik 
latest one -- 1.9.1

Siemienik:
@ns20011997 it is probably a similar issue like here: https://github.com/exceljs/exceljs/pull/722

if I am right, this workaround should works:
```javascript
cell.style = JSON.parse(JSON.stringify(cell.style)); //very bad way to clone anythink, but works.
cell.fill = fill
```

vegarringdal:
hmm, is this because he reads a file and make edits?
Never seen this issue, but maybe because I make the excel file/sheet.

DantSu:
Steps : 
- create a file
- open it with MS excel
- select cells A1 to C5 (It's an example)
- assign a style
- close the file
- open it with exceljs
- run `ws.getCell('B2').font = {...cell.font, color: {argb: 'FF3366FF'}}`
- style of A1 to C5 will be changed

### Workaround

```javascript
const cell = ws.getCell('B2')
cell.style = {
   ...cell.style,
   font: {...cell.font, color: {argb: 'FF3366FF'}}
}
```

### Idea

May be, create a method `setStyle()` on Cell class.

```javascript
setStyle(style) {
  this.style = {...this.style, ...style}
}
```

Siemienik:
@DantSu did you opened PR?

DantSu:
Hi,
If you find that is a good idea too. I will open a pull request.

LeanSaucecode:
> Steps :
> 
> * create a file
> * open it with MS excel
> * select cells A1 to C5 (It's an example)
> * assign a style
> * close the file
> * open it with exceljs
> * run `ws.getCell('B2').font = {...cell.font, color: {argb: 'FF3366FF'}}`
> * style of A1 to C5 will be changed
> 
> ### Workaround
> ```js
> const cell = ws.getCell('B2')
> cell.style = {
>    ...cell.style,
>    font: {...cell.font, color: {argb: 'FF3366FF'}}
> }
> ```
> 
> ### Idea
> May be, create a method `setStyle()` on Cell class.
> 
> ```js
> setStyle(style) {
>   this.style = {...this.style, ...style}
> }
> ```
> 
> #### Use example :
> ```js
> const cell = ws.getCell('B2')
> cell.style = {
>    ...cell.style,
[...]

PR #1378, opened 2020-07-08
fix issue #791 - modify a cell style on existing excel file may affec…

DantSu:
Steps : 
- create a file
- open it with MS excel
- select cells A1 to C5 (It's an example)
- assign a style
- close the file
- open it with exceljs
- run `ws.getCell('B2').font = {...cell.font, color: {argb: 'FF3366FF'}}`
- style of A1 to C5 will be changed

see issue #791 for more informations

### Workaround

```javascript
const cell = ws.getCell('B2')
cell.style = {
   ...cell.style,
   font: {...cell.font, color: {argb: 'FF3366FF'}}
}
```

## Test plan

Added setStyle method to Cell class : 

```javascript
setStyle(style) {
  this.style = {...this.style, ...style}
}
```

### Example 1 : 

```javascript
const cell = ws.getCell('B2')
cell.style = {
   ...cell.style,
   font: {...cell.font, color: {argb: 'FF3366FF'}}
}
```

become : 

```javascript
const cell = ws.getCell('B2')
cell.setStyle({font: {...cell.font, color: {argb: 'FF3366FF'}}})
```

### Example 2 : 

```javascript
const cell = ws.getCell('B2')
cell.style = {
  ...cell.style,
  fill: {
    type: 'pattern',
    pattern: 'solid',
    fgColor: {argb: 'FFD8D8D8'}
  }
}
```

become : 

```javascript
ws.getCell('B2').setStyle({
  fill: {
    type: 'pattern',
    pattern: 'solid',
    fgColor: {argb: 'FFD8D8D8'}
  }
})
```


DantSu:
I haven't update the style section of README file. Do I have to do this ?

Siemienik:
@DantSu Thank you for contributing ExcelJS, it is always welcome for us :)
However I suppose that this PR do not resolve any problem in current logic, it is just adding new function with walk-around . Behaviour of `cell.font` and `cell.border` will stay still unpredictable when it works on read file.

DantSu:
What do you suggest to fix this issue ?

Siemienik:
I suggest that, style object should be immutable - any change on cell.font, cell.border, cell.style should make copy with new value set. Like {...style, font:newFont}

DantSu:
Immutable style object is definitely the best way. But it's a breaking change. The current documentation is based on mutable object.

Siemienik:
> Immutable style object is definitely the best way. But it's a breaking change. The current documentation is based on mutable object.

true :/

darron1217:
This is just adding a helper function but it may help us a lot.

How about merging it as workaround and then create another Issue/PR to apply `immutable` gradually?

Issue #1363, opened 2020-06-30
[F] + Possible Implementation - AutoFilter: Support the filterColumn tag


remzh:
## 🚀 Feature Proposal

Although AutoFilter support has been implemented since 2017, there is currently no way to exclude specific columns within the range. While this works completely fine in most scenarios, it primarily causes trouble with merged cells - as a filter will show up under each individual cell despite the "merge" (which we may not want). 

Adding an AutoFilter with merged cells with Excel:
![image](https://user-images.githubusercontent.com/20567781/86170086-da371780-bad7-11ea-886c-979c6dbd087f.png)

Adding an AutoFilter with merged cells with ExcelJS:
![image](https://user-images.githubusercontent.com/20567781/86170171-f3d85f00-bad7-11ea-9553-91107d6e3b9f.png)

## Motivation/Use Cases
By supporting column exclusions, users can further customize how autoFilter is implemented as well as properly handling merged cells (resulting in the top image instead of the bottom image). 

## Example/Possible Implementation 

From comparing the xlsx files between ones created by ExcelJS and Excel, it appears that ExcelJS is simply missing support for the filterColumn tag. 

```XML
<autoFilter ref="A1:G1"><filterColumn colId="1" showButton="0"/><filterColumn colId="2" showButton="0"/><filterColumn colId="3" showButton="0"/><filterColumn colId="5" showButton="0"/></autoFilter>
```

I've tweaked ExcelJS locally to work around this issue by adding an `exclude` property under `worksheet.autofilter`, so columns can be excluded like this: 
```JavaScript
sheet.autoFilter = {
    from: 'A10', 
    to: 'G10', 
    exclude: [1, 2, 3, 5]
}; 
```

If this solution looks viable, I'd be more than happy to submit it as a pull request along with updating the documentation to reflect it, just let me know!

NiklasPor:
Is this closed because it's **completed**, because it won't be implemented or because or for what reason? I can't find anything in the docs or in the TS types.

PR #1383, opened 2020-07-09
[F] #1363 - Add support for excluding AutoFilter columns

Ryan778:
## Summary

I brought this up in #1363 several days back. Basically, Excel allows you to exclude specific columns from showing the AutoFilter button when setting an AutoFilter range, which is extremely helpful when you want a field to span more than one column and/or with merged cells. This would implement support for that in ExcelJS. 

## Test plan

This implementation is intended to be as minimally invasive and non-destructive as possible - it merely adds on a property, `exclude`, that can be (optionally) passed onto the `autoFilter` property. The readme has also been updated with an example of this. 

If the user opts not to use said `exclude` property, then autoFilter will behave exactly the same way as before. 

Example: 
```javascript
const sheet = workbook.addWorksheet('My Sheet');

sheet.autoFilter = {
    from: 'A1', 
    to: 'E1', 
    exclude: [1, 2] // excludes columns B and C from showing the AutoFilter button
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -656,6 +656,17 @@ worksheet.autoFilter = {
 }
 ```
 
+Additionally, specific columns can be excluded from showing an auto filter button using `exclude`. The `exclude` property takes an array of columns to exclude (with 0 being the leftmost column), and hides the filter button on said columns. 
+
+```javascript
+// Set an auto filter from A1 to E1, while hiding the filter button on the A and C columns
+worksheet.autoFilter = {
+  from: 'A1',
+  to: 'E1',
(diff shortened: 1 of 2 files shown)


Ryan778:
Additionally, there was a typo in one of the changelogs - a link referencing #306 in the changelog actually linked to #304. 
https://github.com/exceljs/exceljs/pull/1383/commits/9076e4d9c2c88defcdfb52049414d2249afdee12 simply fixes that.

WeslleyDeSouza:
any update on this?

Issue #446, opened 2017-11-29
Is it not possible to add image using the WorkbookWriter?


mfahmirukman:
Previously I tried using the excel file writing method. Works fine, but as the data grows so does the memory consumption. Hence the changing of writing method to streaming method.

Tried doing this on WorkbookWriter

```
var sheet = workbook.addWorksheet('Data');
        var base64Data = req.body.img.toString().replace(/^data:image\/png;base64,/, "");
        fs.writeFile('Tagihan/img.png', base64Data, 'base64', function(err) {
            console.log(err);
        })
var img1 = workbook.addImage({
            filename: 'Tagihan/img.png',
            extension: 'png'
        });

        sheet.addImage(img1, 'A1: H25');
```

got this error

`TypeError: workbook.addImage is not a function`

ritesh-nitw:
This issue is a major release blocker. Surprised that, its not being addressed.. 
Right now, its impossible to use addImage using the WorkbookWriter. Any work around won't work.

ritesh-nitw:
Are you planning to support this in future?

Siemienik:
@mfahmirukman @ritesh-nitw 
I am really sad to say that, but probably here is bug with handling `.png` file.

@mfahmirukman could you test this one with `.jpg` and give me a confirmation?

mfahmirukman:
I'm sorry that I can't help anymore because I'm currently not working on the NodeJS project anymore

kkor:
@Siemienik does closing this issue mean you plan on never supporting images in the workbook writer? Or had this been fixed?

mtawara:
I tried using both png and jpg. Both are not working.

chdh:
Duplicate of #322

Issue #351, opened 2017-06-28
addImage streaming


maksimlikharev:
Is there a plan to add image support to streaming mode?

Imperat:
It seems that there isn't. But we can contribute it!

ritesh-nitw:
@maksimlikharev : Were you able to addImage on stream mode?

maksimlikharev:
was sidetracked to a different project, not doing any exceljs at the moment, don't have time yet for this.

lanbin45:
> @maksimlikharev : Were you able to addImage on stream mode?

Hi, I've seen your reply on other issues of this repo for addImage on stream mode, I wanna know whether you have fixed your problem?

chdh:
Duplicate of #322

Issue #322, opened 2017-05-23
workbook.addImage is not a function


ishantiw:
I have installed latest version of exceljs, however I'm still getting error `workbook.addImage is not a function` when i'm trying to add any image to the workbook. 
PS: I have 0.4.10 version

ishantiw:
I am using Excel.stream.xlsx.WorkbookWriter() to create a workbook writer stream

ishantiw:
Problem is when we create a workbook using writer stream then it doesn't find function `addImage()`. However you can also use the following to add an image.

```
let workbook = Excel.stream.xlsx.WorkbookWriter()
workbook.media.push({
    filename: './Picture1.png',
    extension: 'png',
  })
worksheet.addImage(0, 'B1:E6')
// where 0 is the index of the image you added in workbook, basically its an index of the image you pushed to media array
```
Otherwise simply use below to create a worksheet and it works,
`let workbook = new Excel.Workbook()`

bala-drg:
@ishantiw i'm using the above code and still i'm getting "workbook.addImage is not a function".here is the code

var workbook = Excel.stream.xlsx.WorkbookWriter();
workbook.media.push({
    filename: './Picture1.png',
    extension: 'png',
  })
var worksheet = workbook.addWorksheet();
worksheet.addImage(0, 'B1:E6')

ishantiw:
@guyonroche If you will create a worksheet from `stream.xlsx.WorkbookWriter()` then you will have to use 
`workbook.media.push({
    filename: './Picture1.png',
    extension: 'png',
  })`  
You can get these images using index `workbook.media[0]`. This is just a hack and its better not to use this one.

Recommended usage,
`let workbook = new Excel.Workbook()` to create a workbook and then `workbook.addImage` will work. Cheers!

qiandongyq:
@ishantiw I follow the docs but add images are not working
```
let workbook = new Excel.Workbook()
const imageId1 = workbook.addImage({
filename: './Picture1.png',
extension: 'png',
});
worksheet.addBackgroundImage(imageId1);
```
error: imageId1 returns 0

Please help

ishantiw:
@qiandongyq 
[...]

PR #1405, opened 2020-07-23
README: images not supported in streaming mode

chdh:
Images are not supported in streaming mode. The documentation should indicate this.
See #322, #351, #446


alubbe (approved):
Thanks!

PR #1443, opened 2020-08-29
AddRow/s and InsertRow/s now returning the newly added rows

Subhajitdas298:
## Summary

AddRow/s and InsertRow/s now returning the newly added rows.
Added getRows to return multiple rows in one call

## Test plan

Updated test cases. Small change.


Subhajitdas298:
@guyonroche Thank you for accepting the changes. 
If it is ok, then can you release a new version?

PR #1448, opened 2020-09-08
Add addImage method validation

Yangeok:
**addImage** does not contain validation, so I added it.


Siemienik (commented):
Could you please a test to keep confirming that the approach will work forever

PR #1458, opened 2020-09-12
browser: use TextDecoder and TextEncoder for buffer

myfreeer:
## Summary

Doing a profiling in chrome dev tools shows that the `Buffer.toString()` and `Buffer.from(string)` is using unexpected long cpu time. With the native TextDecoder and TextEncoder it can get much faster in browsers supporting it.
On browsers not supporting TextDecoder, like Internet Explorer, `Buffer.toString()` and `Buffer.from(string)` would not be changed.

References:
https://github.com/feross/buffer/issues/268
https://github.com/feross/buffer/issues/60
https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder
https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder

## Test plan

Profiling below is based on reading `spec/integration/data/huge.xlsx`.

<details> 
  <summary>Profiling result before patch</summary>

before
[Profile-before.json.gz](https://github.com/exceljs/exceljs/files/5212581/Profile-before.json.gz)
</details>

<details> 
  <summary>Profiling result after patch</summary>

after
[Profile-after.json.gz](https://github.com/exceljs/exceljs/files/5212584/Profile-after.json.gz)

</details>

### DIFF
--- a/lib/exceljs.bare.js
+++ b/lib/exceljs.bare.js
@@ -1,4 +1,6 @@
 // this bundle is built without polyfill leaving apps the freedom to add their own
+require('./utils/browser-patch-buffer');
+
 const ExcelJS = {
   Workbook: require('./doc/workbook'),
 };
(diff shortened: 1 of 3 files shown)


alubbe:
This is pretty cool, but I don't like that we're monkey-patching a global dependency. Can't we just use TextDecoder and TextEncoder locally in our code base instead of Buffer.from?

PR #1466, opened 2020-09-17
grunt: skip babel transpile for core-js

myfreeer:
## Summary

Previously grunt build would transpile the output of browserify to transpile all dependencies, but this would make some dependencies, such as core-js, unexpectedly transpiled. Also, this way of transpiling would create source maps only targeting the browserify output instead of original source, and makes the final `dist/exceljs.js` larger than expected.

* Fix Symbol.toString() related errors on IE11 caused by transpiling core-js
* Reduce size for compiled dist
* Keep original source maps before browserify

## Test plan

Existing ci builds and manually tested in browser.
<details>
<summary>screenshot of source map</summary>

source_map

</details>


alubbe (approved):
This is great, thanks!

Issue #742, opened 2019-02-08
Low Performance.


IgnatovDan:
Hello,

Currently I am reviewing your library to see whether or not I can use it instead of my own implementation.

For now, I found the only issue with performance:
* [6-7s with our implementation, click the 'Click me to start DataGrid ExcelExport and wait 5-10 secs (there is no visual progress)' button](https://codepen.io/DanIgnatov/pen/qgxbXd)  to export [20000 x 50] array
* [15-20s with ExcelJS, click the 'Click me to start ExcelExport and wait 15-20 secs (there is no visual progress)' button](https://codepen.io/DanIgnatov/pen/vbdLjx?editors=1010) to export [20000 x 50] array

Click the [Click me..] button on each page and review the elapsed time in the shown alert dialog.

Do you have plans to improve it? There are no notes about performance in the [Backlog](https://github.com/exceljs/exceljs#backlog)

Siemienik:
@IgnatovDan thank you for test, could I please You to give me repo with benchmark ?

IgnatovDan:
Thank you for the reply.
I would like to help you to improve this scenario but it is not clear what is the 'repo with benchmark'?

Right now I created two simple pages at codepen.io (see my first comment):
* [6-7s with our implementation, click the 'Click me to start DataGrid ExcelExport and wait 5-10 secs (there are no visual progress)' button](https://codepen.io/DanIgnatov/pen/qgxbXd)  to export [20000 x 50] array
* [15-20s with ExcelJS, click the 'Click me to start DataGrid ExcelExport and wait 15-20 secs (there are no visual progress)' button](https://codepen.io/DanIgnatov/pen/vbdLjx?editors=1010) to export [20000 x 50] array

Click the [Click me..] button on each page and review the elapsed time in the shown alert dialog. I have simplified these pages and now there is the only button that shows the elapsed time in an alert dialog.

alubbe:
Just did a quick check - datagrid seems to use a lower zip compression level, so exceljs is actually doing a lot more work. But I feel like when generating xlsx files on the client, we should favour cpu cycles over saving output size (4.2 MB vs 4.4 MB in this case). Maybe we can change the default compression to the lowest and fastet level and then provide an option to increase/change it for those who need it

alubbe:
Okay so never mind, that's not it after all. The zip compression is the same, there must be something very inefficient in `workbook.xlsx.writeBuffer()` after all

rogozind:
[...]

PR #1476, opened 2020-09-28
Increase the performance of some xml and html helpers

alubbe:
## Summary

We saw some performance issues in https://github.com/exceljs/exceljs/issues/742 that this PR tries to solve

## Test plan

Both helpers are already tested, this just refactors their implementation.

### DIFF
--- a/lib/utils/under-dash.js
+++ b/lib/utils/under-dash.js
@@ -1,5 +1,5 @@
 const {toString} = Object.prototype;
-
+const escapeHtmlRegex = /["&<>]/;
 const _ = {
   each: function each(obj, cb) {
     if (obj) {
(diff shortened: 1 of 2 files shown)


Siemienik (approved):
Thank you for this PR, nice done
How big speedup you note after these changes?

lib/utils/under-dash.js
```diff
@@ -84,12 +84,39 @@ const _ = {
...
-      .replace(/'/g, '&#039;');
+    const regexResult = escapeHtmlRegex.exec(html);
+    if (!regexResult) return html;
+
+    let result = '';
+    let escape = '';
+    let lastIndex = 0;
+    let i = regexResult.index;
+    for (; i < html.length; i++) {
+      switch (html.charCodeAt(i)) {
```
rogozind, line 95: For readability I would use html.charAt() and the actual characters in case ('>' vs 62). I tested both variants and performance is identical.

rogozind:
This is good. Regexp way of finding whether or not we need the escaping is even faster than the loop I suggested #742 .
In my test I got 25% increase in performance. I have 43K rows of data.  It takes 20 secs to export. 50% used to go to create the spreadsheet and 50% used to go to write out the files. Out of these 10 secs 5 secs went to serialize various elements and 5 secs went to xmlEncode() calls. With the patch the xmlEncode dropped to near zero so export now takes 15secs.

Please accept this PR.

lib/utils/under-dash.js
alubbe: Good catch!

PR #1478, opened 2020-09-29
Fix #1364 Incorrect Worksheet Name on Streaming XLSX Reader

### LINKED ISSUE #1364: [BUG] Incorrect Worksheet Name on Streaming XLSX Reader
## 🐛 Bug Report

I am having problems getting correct worksheet name when using streaming xlsx reader (async iteration). The first worksheet always defaults to *Sheet1* regardless of name.

Lib version: 4.1.0

excel
sheet-name

## Steps To Reproduce

[Sample Template.xlsx](https://github.com/exceljs/exceljs/files/4855135/Sample.Template.xlsx)

```javascript
const workbook = new stream.xlsx.WorkbookReader('path/to/Sample Template.xlsx', {});

for await (const worksheetReader of workbook) {
    console.log('sheet name:', worksheetReader.name); // It always outputs Sheet1
}
```

## The expected behaviour:

Worksheet name should be _Sum Worksheet_ instead of Sheet1.

antimatter15:
## Summary

This PR fixes issue #1364 where the streaming XLSX reader emits incorrect sheet names.

## Test plan

The sample template from the issue is added as an integration test.


alubbe (approved):
Thank you!

Issue #1109, opened 2020-02-04
how to set validation on a column


swatantra1:
i want to generate an excel for students where i have a section gender and i want the column to have validation as list = ["male", "female", "other"] 
how can i do so

Siemienik:
https://github.com/exceljs/exceljs#data-validations

bhavan0:
> https://github.com/exceljs/exceljs#data-validations
> 
> ```js
> ws.getColumn("A").eachCell({ includeEmpty: true }, function(cell, rowNumber) {
>   cell.dataValidation = {
>     type: 'list',
>     allowBlank: true,
>     formulae: ['"male,female,other"']
>   };
> });
> ```

Hi @Siemienik ,

using the above code the validations are only getting added to the column cells which have data during download.

How do we add the validation to the entire column when there is data entered or not?

Siemienik:
#1184

Siemienik:
@bhavan0 may you confirm is it working properly?
```js
ws.dataValidations.add("X2:X9999",{
    type: 'list',
    allowBlank: true,
    formulae: ['"male,female,other"']
  });
```

bhavan0:
> @bhavan0 may you confirm is it working properly?
> 
> ```js
> ws.dataValidations.add("X2:X9999",{
>     type: 'list',
>     allowBlank: true,
>     formulae: ['"male,female,other"']
>   });
> ```

This works.
Thank You :)

Siemienik:
:+1:

maikknebel:
Is this still working for someone? 
ws.dataValidations is shown as an unresolved variable to me. Calling `ws.dataValidations.add(...)` leads to `TypeError: Cannot set property 'marked' of undefined`

jrbarnard:
> Is this still working for someone?
> ws.dataValidations is shown as an unresolved variable to me. Calling `ws.dataValidations.add(...)` leads to `TypeError: Cannot set property 'marked' of undefined`

Seeing the same error

GowthamGottimukkala:
> Is this still working for someone?
> ws.dataValidations is shown as an unresolved variable to me. Calling `ws.dataValidations.add(...)` leads to `TypeError: Cannot set property 'marked' of undefined`

Same error. Someone solved this?

GowthamGottimukkala:
#1027

adnaanbheda:
@GowthamGottimukkala did you find a solution to the `TypeError` ?

[...]

Issue #1027, opened 2019-11-15
Upgraded to 3.4 - Cannot set property 'marked' of undefined


trueter:
Downgrading to 3.3.1 avoids the problem. 
Difficult for me to say what exactly causes the problem as the error is raised only when writing the sheet to file.
```
(node:57552) UnhandledPromiseRejectionWarning: TypeError: Cannot set property 'marked' of undefined
    at .../node_modules/exceljs/lib/xlsx/xform/sheet/data-validations-xform.js:77:40
    at Array.map (<anonymous>)
    at .../node_modules/exceljs/lib/xlsx/xform/sheet/data-validations-xform.js:54:6)
    at .../node_modules/exceljs/lib/xlsx/xform/sheet/data-validations-xform.js:105:28)
    at .../node_modules/exceljs/lib/xlsx/xform/sheet/worksheet-xform.js:233:30)
    at .../node_modules/exceljs/lib/xlsx/xlsx.js:532:22
    at Array.forEach (<anonymous>)
    at .../node_modules/exceljs/lib/xlsx/xlsx.js:530:22)
    at .../node_modules/exceljs/lib/xlsx/xlsx.js:623:16)

```

dadlian:
Thanks man this saved me from a nightmare meeting with a client!

cooperbrown9:
How did you solve it?

andersevenrud:
I had to downgrade to 3.3.1 in order for this to work again. The latest 3.7 has the same issue.

Bene-Graham:
I am also getting this issue.

It seems to occur when I do the following
```js
worksheet.dataValidations.model["B2:B1048576"] = genderValidation;
```

In versions before 3.4 it would apply the validation to all cell in that range. In version 3.4 or greater I get an error.

If i manually iterate through each cell,  it will work, but is very slow compared to how it was before.

Siemienik:
hmm.. it's a quite dizzy situation :) We haven't documentation about `worksheet.dataValidations.model` 

I've found this PR which introduce this feature: #907 

If anyone wanna add a description to docs I will be glad

Siemienik:
I've some investigation about that and for me, these changes are best candidates: https://github.com/exceljs/exceljs/compare/v3.3.1...v3.4.0#diff-883923368e51231ee142f6d7e9396b8bR216

PR #1480, opened 2020-09-30
Fixed undefined ref error when setting a data validation that is a range of cells at the worksheet level

Bene-Graham:
Fixed undefined ref error when setting a data validation that is a range of cells at the worksheet level

## Summary

In the case where a data validation is set at the worksheet level with a range of cells "A1:A100" the error `Cannot set property 'marked' of undefined` would occur.

From what I could tell the original code would call `const addr = colCache.decodeAddress(dv.address);` this would return the value "A1" in the case of the following "A1:A100". Later on in the code it would try to mark "A1" `dvMap[otherAddress].marked = true;`, but in the dvMap it would be "A1:A100" and the error occurs.

I changed decodeAddress to decodeEx so that we can detect if the address is a single cell vs a range(dimension?). If a  dimensions property exist on the return object from decodeEx, it will shortcircuit and use the range.

## Test plan
I added a integration test demonstrating it now works with no error: `spec/integration/issues/issue-1027-sheet-defined-data-validate.spec.js`

### DIFF
--- a/lib/xlsx/xform/sheet/data-validations-xform.js
+++ b/lib/xlsx/xform/sheet/data-validations-xform.js
@@ -51,7 +51,15 @@ function optimiseDataValidations(model) {
   return dvList
     .map(dv => {
       if (!dv.marked) {
-        const addr = colCache.decodeAddress(dv.address);
+        const addr = colCache.decodeEx(dv.address);
+        if (addr.dimensions) {
+          // If we comeback with a dimensions, we can assume this was sone on purpose and should not be optimized
+          dvMap[addr.dimensions].marked = true;
+          return {
+            ...dv.dataValidation,
+            sqref: dv.address,
+          };
+        }
 
         // iterate downwards - finding matching cells
         let height = 1;
--- a/lib/xlsx/xform/sheet/data-validations-xform.js
+++ b/lib/xlsx/xform/sheet/data-validations-xform.js
@@ -53,7 +53,6 @@ function optimiseDataValidations(model) {
       if (!dv.marked) {
         const addr = colCache.decodeEx(dv.address);
         if (addr.dimensions) {
-          // If we comeback with a dimensions, we can assume this was sone on purpose and should not be optimized
           dvMap[addr.dimensions].marked = true;
           return {
             ...dv.dataValidation,
--- a/spec/integration/issues/issue-1027-sheet-defined-data-validate.spec.js
+++ b/spec/integration/issues/issue-1027-sheet-defined-data-validate.spec.js
@@ -1,0 +1,27 @@
+const ExcelJS = verquire('exceljs');
+
+// this file to contain integration tests created from github issues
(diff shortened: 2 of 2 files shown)


Bene-Graham:
One additional thing I believe this will solve the following issue: #1109 and #1027

spec/integration/issues/issue-1027-sheet-defined-data-validate.spec.js
```diff
@@ -0,0 +1,27 @@
+const ExcelJS = verquire('exceljs');
+
+// this file to contain integration tests created from github issues
```
alubbe, line 3: please remove this line

spec/integration/issues/issue-1027-sheet-defined-data-validate.spec.js
```diff
@@ -0,0 +1,27 @@
...
+      errorTitle: 'Invalid Selection',
+      formulae: ['"Apples,Bananas,Oranges"'],
+      showErrorMessage: true,
+      type: 'list',
+    };
+
+    return wb.xlsx.writeFile(TEST_XLSX_FILE_NAME).then(() => {
+      // arriving here is success
+      expect(true).to.equal(true);
+    });
```
alubbe, line 25: Please change this to just 
```js
await wb.xlsx.writeFile(TEST_XLSX_FILE_NAME);
// arriving here is success
```

lib/xlsx/xform/sheet/data-validations-xform.js
```diff
@@ -51,7 +51,15 @@ function optimiseDataValidations(model) {
   return dvList
     .map(dv => {
       if (!dv.marked) {
-        const addr = colCache.decodeAddress(dv.address);
+        const addr = colCache.decodeEx(dv.address);
+        if (addr.dimensions) {
+          // If we comeback with a dimensions, we can assume this was sone on purpose and should not be optimized
```
alubbe, line 56: I think it's fine to remove this comment

alubbe (changes requested):
Thank you for this PR, I just have some minor modifications that I'd like to see and then we're good to go

spec/integration/issues/issue-1027-sheet-defined-data-validate.spec.js
Bene-Graham: I have removed the comment line

spec/integration/issues/issue-1027-sheet-defined-data-validate.spec.js
Bene-Graham: Since none of the other test are using await / async, I ended up just doing: `return wb.xlsx.writeFile(TEST_XLSX_FILE_NAME);`

lib/xlsx/xform/sheet/data-validations-xform.js
Bene-Graham: Done

alubbe (approved):
thank you!

PR #1481, opened 2020-10-01
Fix type definition for DefinedNamesRanges

antimatter15:
## Summary

The `DefinedNamesRanges` type actually has a `ranges` property, not a `range` property. 

## Test plan

This is the code that populates the named ranges list: https://github.com/exceljs/exceljs/blob/master/lib/doc/defined-names.js#L143

As you can tell the field is named `ranges`, not `range`


alubbe (approved):
thank you

PR #1482, opened 2020-10-01
Performance improvement in col-cache

antimatter15:
## Summary

Certain XLSX files have references to "TRUE" and "FALSE" that get ultimately passed over to the col-cache l2n method. When l2n is passed a string of 4 characters or more, it ends up regenerating the dictionary from scratch each time.

For a particular XLSX file I was working with, this change resulted in an overall 2x performance improvement. 

## Test plan

This change doesn't affect behavior, as `l2n` already throws when it encounters a string which is out of range. This change makes it throw earlier- before repopulating a 16384-length array each time.

### DIFF
--- a/lib/utils/col-cache.js
+++ b/lib/utils/col-cache.js
@@ -47,6 +47,9 @@ const colCache = {
     let l2;
     let l3;
     let n = 1;
+    if (level >= 4) {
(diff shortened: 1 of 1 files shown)


alubbe (approved):
good catch, thank you!

Issue #1406, opened 2020-07-24
[F] The printing size can be set to A3


skypesky:
## 🚀 Feature Proposal

The expected papersize support for printing is set to A3. (pageSetup.paperSize = Number corresponding to A3)

## Motivation

Users will be able to set papersize to A3.

## Example

```typescript
const workbook: Workbook = new Workbook();
const worksheet: Worksheet = workbook.addWorksheet(sheetName, {
    pageSetup: {
        // Print size set to A3
        paperSize: Number corresponding to A3,
    },
});
```

skypesky:
Solution
```ts
const workbook: Workbook = new Workbook();
const worksheet: Worksheet = workbook.addWorksheet(sheetName, {
    pageSetup: {
        // Print size set to A3
        paperSize: 8,
    },
});
```

PR #1483, opened 2020-10-01
add A3 paperSize number

skypesky:
## Summary

Add the corresponding papersize for A3 to improve the document.
I didn't see any unit tests about papersize, so I didn't add unit tests, but I think Microsoft's excel file specifications are the same.
I only tested the papersize of A3. [Reference documents](https://docs.microsoft.com/zh-cn/office/vba/api/excel.xlpapersize)


skypesky:
#1406 The origin of the problem

Issue #1474, opened 2020-09-23
[BUG] Incorrectly handles '/', ':' characters in sheet name


xydens:
## 🐛 Bug Report

If the desired name of the worksheet contains characters `/`, `:` it does not affect on actual sheet name. When I open created file via MS Excel is see default worksheet names like 'Sheet 1'. Some versions of MS Excel also show error messages and warn about file corruption. 

Lib version: 4.1.1

## Steps To Reproduce

1. Have `exceljs@4.1.1` installed (obviously)
2. Run following code:

```javascript
const {Workbook} = require('exceljs')

const workbook = new Workbook()
const sheet = workbook.addWorksheet('http://a.b/c')
sheet.addRow(['a', 'b', 'c'])
workbook.xlsx.writeFile('./result.xlsx')
```
3. See 'Sheet 1' instead of desired sheet name in created file

## The expected behaviour:

Sheet name to be equal to first argument of `addWorksheet` call.

## Possible solution (optional, but very helpful):

If it is xlsx constraint, throw an error, when client attempts to put unsupported characters into sheet name.

skypesky:
@xydens To solve this problem, I have submitted a PR. you can update exceljs to the latest version.

xydens:
Thank you!

PR #1484, opened 2020-10-02
Fix issue 1474 (to check invalid sheet name)

skypesky:
## Summary

Fix issue #1474 .
An exception is thrown when the name is illegal.

## Test plan

For this I have added unit tests.


papb:
Hello! I suggest you take a look [at this](https://github.com/papb/json-excel/blob/master/source/assert-valid-sheet-names.ts), do you think it would be good to also add these other checks?

skypesky:
hi,papb,Your method is more concise, I think it's very good.
I didn't write according to the assertvalidsheetname method before, because the validation of name in the original code did not always throw an exception. In this way, how to organize it becomes very strange.

About double quotation marks
---
It seems that the naming rules of sheet have something to do with the version of Excel

As for single quotation marks
---
As far as I know, single quotation marks can be placed in the middle, but not the head and tail. Can you test the single quotation marks for me?

spec/integration/worksheet.spec.js
```diff
@@ -659,6 +659,22 @@ describe('Worksheet', () => {
       });
     });
 
+    context('when the worksheet name contains illegal characters', () => {
+      it('when the worksheet name contain < > * ? " | : / \\ [ ]', () => {
```
alubbe, line 663: this should be `it('throws an error'...)`

spec/integration/worksheet.spec.js
```diff
@@ -659,6 +659,22 @@ describe('Worksheet', () => {
...
+        const workbook = new ExcelJS.Workbook();
+
+        const invalideNames = ['*', '?', ':', '/', '\\', '[', ']'];
```
alubbe, line 666: please call this `invalidCharacters`

lib/doc/workbook.js
```diff
@@ -97,6 +102,12 @@ class Workbook {
     return worksheet;
   }
 
+  // Illegal character in worksheet name: asterisk (*), question mark (?),
+  // colon (:), forward slash (/ \), or bracket ([])
+  containIllegalSheetName(sheetName) {
+    return /(\*)|(\?)|(:)|(\/)|(\\)|(\[)|(\])/.test(sheetName);
```
alubbe, line 108: please use the simpler and faster regex `/[*?:\/\\\[\]]/` instead

alubbe (changes requested):

spec/integration/worksheet.spec.js
```diff
@@ -659,6 +659,48 @@ describe('Worksheet', () => {
...
+            workbook.addWorksheet(invalidCharacters[lastIndex])
+          ).to.throw(expectedErrors[lastIndex]);
+        }
+      });
+
+      it('throws an error', () => {
+        const workbook = new ExcelJS.Workbook();
+
+        const invalidNames = ['\'sheetName', 'sheetName\''];
+        const expectedErrors = invalidNames.map(
```
alubbe, line 687: I think the array creation and the for-index-loop distract a little bit from what is being tested, I think we should simplify to this:
```js
for (const invalidName of invalidNames) {
  expect(() => workbook.addWorksheet(invalidName)).to.throw(
    `The first or last character of worksheet name cannot be a single quotation mark: ${invalidName}`
  );
}
```

spec/integration/worksheet.spec.js
```diff
@@ -659,6 +659,48 @@ describe('Worksheet', () => {
       });
     });
 
+    context('when the worksheet name contains illegal characters', () => {
+      it('throws an error', () => {
+        const workbook = new ExcelJS.Workbook();
+
+        const invalidCharacters = ['*', '?', ':', '/', '\\', '[', ']'];
+        const expectedErrors = invalidCharacters.map(
```
alubbe, line 667: I think the array creation and the for-index-loop distract a little bit from what is being tested, I think we should simplify to this:
```js
for (const invalidCharacter of invalidCharacters) {
  expect(() => workbook.addWorksheet(invalidCharacter)).to.throw(
    `Worksheet name ${invalidCharacter} cannot include any of the following characters: * ? : \\ / [ ]`
  );
}
```

alubbe (changes requested):
Just some minor feedback around the readability/maintainability of the test case, then we're good to go!

alubbe (approved):
thank you!

Sczlog:
Although here has been a long long times since this change applied.
But I suppose if replace invalid character with `_` should be a better choice?
Current solution is kinda a breaking change, code running successful (but not correct?) before will break after update to later version.

PR #1487, opened 2020-10-04
Add lastColumn property (fixes #1453)

### LINKED ISSUE #1453: Add type for property of worsheet.lastcolumn
## 🚀 Feature Proposal
Similar to worsheet.lastRow worksheet.lastColumn should be added 

## Motivation
Doing so will facilitate the creation of multiple tables in same sheet and adding address for next table

## Example

Please provide an example for how this feature would be used.

FliegendeWurst:
## Summary

As suggested by @Jigna-Rathod in #1453, a `lastColumn` property is added in this PR.

## Test plan

I created an automated test in `gold.spec.js`.

### DIFF
--- a/lib/doc/worksheet.js
+++ b/lib/doc/worksheet.js
@@ -260,6 +260,10 @@ class Worksheet {
     this.workbook.definedNames.spliceColumns(this.name, start, count, inserts.length);
   }
 
+  get lastColumn() {
+    return this.getColumn(this.columnCount);
+  }
+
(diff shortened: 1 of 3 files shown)


Siemienik (commented):
What happens if file has empty columns? - Could I ask you to add appropriate test?

FliegendeWurst:
I added another test based on the 1904.xslx file (the first column is empty). As expected, lastColumn is the second column.

Issue #1473, opened 2020-09-23
Export CSV garbled characters


jingwuluo:
When emoji is included in worksheet.columns, the encoding format of the exported CSV file will not be the specified encoding format

jingwuluo:
I looked at the source code and found that when using the csv.writeBuffer() method, the parameter encoding has no effect.
![image](https://user-images.githubusercontent.com/30119218/93994062-0c4bd300-fdc2-11ea-844b-865d4829b851.png)

Issue #995, opened 2019-10-24
Can't get hebrew to display correctly in a generted CSV file


adame21:
Hello,

I've recently started using the lib and am loving the ease of use and options.

My only problem is that hebrew strings are not displaying correctly.

I have tried all kinds of encoding/decoding but nothing seems to work.

Does this lib support hebrew? if so, can you point me in the right direction to get it working?

What im doing right now is this: 

````
                worksheet.addRow({
message:"משהו שכתוב בעברית" // this translates to: "something in hebrew" 
})

````
The CSV file gets generated perfectly, but what I get in the "message" field is something like this:
׳׳©׳”׳• ׳©׳›׳×׳•׳‘ ׳‘׳¢׳‘׳¨׳™׳×

I hope this is enough information to resolve this.
Once again - loving this lib, thanks so much for making it.

Thanks :)

**Comment by ArtskydJ** [comment]
I think the problem is with the encoding. When you save it, try something like this:
```js
const wb = new ExcelJS.Workbook();
const ws = wb.addWorksheet('wheee');
ws.getCell('A1').value = 'משהו שכתוב בעברית';
wb.csv.writeFile('test.csv', { encoding: 'UTF-8' })
```

PR #1495, opened 2020-10-09
Add a test for CSV writeFile encoding

ArtskydJ:
## Summary

I saw a few issues (#995, #1473) regarding CSV output encoding, and I wondered if it was working.

## Test plan

So I wrote a test... Turns out it is working.

### DIFF
--- a/spec/integration/issues/issue-995-encoding.spec.js
+++ b/spec/integration/issues/issue-995-encoding.spec.js
@@ -1,0 +1,25 @@
+const ExcelJS = verquire('exceljs');
+
+const TEST_CSV_FILE_NAME = './spec/out/issue-995-encoding.test.csv';
+const HEBREW_TEST_STRING = 'משהו שכתוב בעברית';
+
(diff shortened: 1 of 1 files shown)


alubbe (approved):

pputnam-NAVEX:
Was the new Excel document opened and checked after it was written to? I can successfully encode the file as UTF-8, check in console fine, but when Excel opens the document characters including the Hebrew text used in the test are gibberish. Notepad++ correctly displays it and if I use Notepad++ to encode it to UTF-8-BOM I can re-open the spreadsheet and it will display correctly.

pputnam-NAVEX:
Adding:
worksheet.getCell('A1').value = "\ufeff" + worksheet.getCell('A1').value; 

Right before the writeFile works, also if this same file is then read using readFile \ufeff is undetectable - unsure if this is something worth adding as BOM isn't recommended but limitations in viewability of UTF-8 characters when opening actual Excel files for non-technical or data visualization may be helpful.

Issue #1430, opened 2020-08-17
[BUG] date parsing doesn't work in streaming mode


contra:
## 🐛 Bug Report

Dates are not being parsed correctly when using streaming - issue seems to be the parser internally not getting the "style" for the field.

Lib version: 4.1.1 (latest)

## Steps To Reproduce

```javascript
const through = require('through2')
const { pipeline, Readable } = require('readable-stream')

const reader = new Excel.stream.xlsx.WorkbookReader(fs.createReadStream('file.xlsx'))
const createReader = async function* () {
  for await (const worksheet of reader) {
    for await (const row of worksheet) {
      yield row
    }
  }
}
  

const out = pipeline(
  Readable.from(createReader()),
  through.obj(function (row, _, cb) {
    console.log(row.values)
    cb(null, row)
  }),
  (err) => {
    if (err) out.emit('error', err)
  }
)
```

You'll see the second column always comes back as a number, even though it should be an ISO string. This file is parsed perfectly fine and gives the correct results when not using streaming mode.

Attached the excel file, you can see the second column is properly date formatted.

[file.xlsx](https://github.com/exceljs/exceljs/files/5087163/file.xlsx)

## Possible solution

I debugged into the library, it seems like the issue is that [this](https://github.com/exceljs/exceljs/blob/master/lib/stream/xlsx/worksheet-reader.js#L278) is always returning null and not finding the style, which means that later on when it checks the number format to see if it is a date it returns false and just treats it as a number. In the case of this excel sheet, `c.s` is `2`

contra:
Dump of the XML for this workbook from `utils/parse-sax.js`:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<worksheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main" xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships">
  <sheetPr filterMode="false">
    <pageSetUpPr fitToPage="false" />
  </sheetPr>
  <dimension ref="A1:D65536" />
  <sheetViews>
    <sheetView windowProtection="false" showFormulas="false" showGridLines="true" showRowColHeaders="true" showZeros="true" rightToLeft="false" tabSelected="true" showOutlineSymbols="true" defaultGridColor="true" view="normal" topLeftCell="A1" colorId="64" zoomScale="100" zoomScaleNormal="100" zoomScalePageLayoutView="60" workbookViewId="0">
[...]

PR #1531, opened 2020-11-13
docs: fix default options for streaming xlsx reader

contra:
## Summary

The docs do not match the implementation or the type information: https://github.com/exceljs/exceljs/blob/4e38ad8ed2a18b6fe3999528cd30bf8b6579c249/index.d.ts#L1958

Closes https://github.com/exceljs/exceljs/issues/1430

## Semi-related

- I think it would improve the docs to identify what "styles" means to the user - it isn't clear that type information (dates, for example) are a style and without `styles: 'cache'` all dates are unusable.
- The default values for these options seem like a bit of a footgun, maybe worth revisiting why these are the defaults?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2491,10 +2491,10 @@ The constructor takes a required input argument and an optional options argument
 | --------------------- | ----------- |
 | input (required)      | Specifies the name of the file or the readable stream from which to read the XLSX workbook. |
 | options (optional)    | Specifies how to handle the event types occuring during the read parsing. |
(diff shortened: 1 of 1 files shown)


alubbe:
I think I'm in favour of both changing the defaults and better documenting what the options do, and that you probably only want to disable the caching for a small performance boost. Would you mind adding that to this PR?

contra:
@alubbe Will do.

PR #1544, opened 2020-11-30
Typing Table add row optional rowNumber

Andries-Smit:
## Summary

Using typescript the Table function `addRow` has a `rowNumber` parameter which is required but should be optional.

In the implementation https://github.com/exceljs/exceljs/blob/master/lib/doc/table.js#L341 there is a checks for undefined. And described in the comment above `either insert at rowNumber or append`


alubbe (approved):

alubbe:
Thank you!

PR #1550, opened 2020-12-03
Typing FillPattern fgColor should be optional

Andries-Smit:
## Summary

Using typescript the FillPattern the `fgColor` should be optional.

This is described in the documentation https://github.com/exceljs/exceljs/blame/master/README.md#L1722


Siemienik (approved):
Good catch, thank you. LGTM:+1:

---

PR #1564, opened 2020-12-15
Fixed return type on getRows

pmcilwaine:
## Summary

This PR updates the typescript definition file to have the correct return type for getRows. 

## Test plan

As you can see from this file https://github.com/exceljs/exceljs/blob/d76dc388d6b0914262feff24eb6386d4c030e4bb/lib/doc/worksheet.js#L352 when the range is < 1 it will return undefined.

I have also added a unit test to demonstrate that when the range is < 1 it returns undefined

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1244,7 +1244,7 @@ export interface Worksheet {
 	/**
 	 * Get or create rows by 1-based index
 	 */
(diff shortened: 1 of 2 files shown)


Siemienik (approved):
Thank you for great fix :) LGTM 🥇 

---

PR #1570, opened 2020-12-17
fix a handful of streaming parser bugs

### LINKED ISSUE #1430: [BUG] date parsing doesn't work in streaming mode

### LINKED ISSUE #1558: [bug] iterate-stream causes race conditions in streaming mode
## 🐛 Bug Report

When using async iterables to parse a workbook as a stream, randomly the zip parser hangs due to https://github.com/exceljs/exceljs/blob/master/lib/utils/iterate-stream.js

Just from a quick glance the utility is really non-standard and full of unhandled edge cases.

Lib version: 4.2.0

## Steps To Reproduce

This is extremely difficult to reproduce and I haven't been able to make an isolated test case. I have been loading up 100 parsers concurrently in a loop to reproduce this specific race condition.

This utility isn't needed since all readable streams are async iterable in node core already, this is probably just old code from before this feature landed in core. Simple removing usage of this utility and just using the streams as iterables directly solved the bug. If this all seems sound to you, I can send a PR to remove the utility and switch to the core behavior. I noticed some other behavior that should be cleaned up around temp file management as well that I can address in the same PR.

contra:
## Summary

We use the streaming parser and encountered a number of issues - sometimes files would stall out, sometimes they would always fail unless you opened and resaved the file, sometimes dates would just not be parsed for some files.

### Major Changes

- Change default arguments to match the docs (Closes #1531)
- Fixes a race condition where worksheets would start being read before the styles were parsed, causing dates to fail and show up as numbers (Closes #1430)
- Fixes a stall (unreported?) in the BaseXForm parser - if closetag was encountered it would return, and the sax parser would fail to flush and just pause forever. The correct behavior should be to break so that it flushes the parser correctly.
- Fixes a stall (unreported?) where worksheets would never be read.
  - I resolved this by implementing a basic queue to handle the waitingWorksheets logic - I think this cleans up the code a bit as well and makes it easier to understand.
- Removes the `iterateStream` utility, which was causing stalls and race conditions (Closes #1558)
  - I read through all of the previous conversation on the unzipper project so I understand the initial reason why it was added.
  - All tests pass without it, I think this was just covering up for some of the previous bugs - file parser generators were stalling which bubbles up causing the unzipper generator to stall as well (and thus fail to autodrain, because that statement was never reached).

### Minor Changes

- Changes the getStream ducktype to match the other code in the project (just checking for a pipe function)
- Use pipeline when writing files so that errors are properly managed
- Removed some node 8 code - Node 8 has been deprecated since 2019 (https://blog.risingstack.com/update-nodejs-8-end-of-life-no-support/)
- Switches to the promise version of autodrain, since your PR to unzipper making that work correctly landed and is published

## Test plan

All of our edge case tests that use exceljs under the hood pass, and the existing exceljs tests pass. Let me know if you would like to see any additional tests.


lib/stream/xlsx/workbook-reader.js
```diff
@@ -25,18 +23,20 @@ class WorkbookReader extends EventEmitter {
...
+      hyperlinks: 'cache',
+      styles: 'cache',
+      entries: 'emit',
       ...options,
     };
 
     this.styles = new StyleManager();
     this.styles.init();
+    this._parsedStyles = false;
+    this.waitingWorkSheets = [];
```
Siemienik, line 35: As far as I understood well, It might be a `const` living in the scope of `parse`. Then passed into `queueWorksheets` and `flushQueue`as a parameter. 
I think that will be much safer, and much more hermetic.

lib/stream/xlsx/workbook-reader.js
```diff
@@ -76,16 +76,60 @@ class WorkbookReader extends EventEmitter {
...
+    const queue = this.waitingWorkSheets; // make a copy
+    this.waitingWorkSheets = []; // clear, flushQueue can be called again while this is still processing async
+    for (const {sheetNo, path, tempFileCleanupCallback} of queue) {
+      const fileStream = fs.createReadStream(path);
+      yield* this._parseWorksheet(fileStream, sheetNo);
+      tempFileCleanupCallback();
+    }
+  }
+
+  cleanupQueue() {
```
Siemienik, line 91: This function is never called. Is it required? If not, please delete it.

lib/xlsx/xform/base-xform.js
```diff
@@ -64,9 +64,7 @@ class BaseXform {
         } else if (eventType === 'text') {
           this.parseText(value);
         } else if (eventType === 'closetag') {
-          if (!this.parseClose(value.name)) {
-            return this.model;
-          }
+          if (!this.parseClose(value.name)) break;
```
Siemienik, line 67: which issue is it fix?

lib/xlsx/xlsx.js
```diff
@@ -285,9 +280,11 @@ class XLSX {
           entryName = entryName.substr(1);
         }
         let stream;
-        if (entryName.match(/xl\/media\//) ||
+        if (
+          entryName.match(/xl\/media\//) ||
```
Siemienik, line 284: which issue is it fix?

Siemienik (commented):
Hi @contra, Thank you for the great contribution :) Great job 🥇 

Could I please you to do two points?
1. Why test on MacOS with node v10 & v12 fails?
2. Update this branch to the current master.

As these changes are major (min node version), I suppose we should separate major changes from another. To protect backwards-compatibility.

Additionally, I wan't to merge it without other contribution also approve major changes.

---

lib/stream/xlsx/workbook-reader.js
```diff
@@ -25,18 +23,20 @@ class WorkbookReader extends EventEmitter {
     this.options = {
       worksheets: 'emit',
       sharedStrings: 'cache',
-      hyperlinks: 'ignore',
-      styles: 'ignore',
-      entries: 'ignore',
+      hyperlinks: 'cache',
+      styles: 'cache',
+      entries: 'emit',
```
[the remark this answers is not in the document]
contra: @Siemienik I was asked to do this by @alubbe https://github.com/exceljs/exceljs/pull/1531 - the docs do not currently match the implementation, so this makes the implementation match the docs which is the intended behavior.

lib/xlsx/xform/base-xform.js
contra: "Fixes a stall (unreported?) in the BaseXForm parser - if closetag was encountered it would return, and the sax parser would fail to flush and just pause forever. The correct behavior should be to break so that it flushes the parser correctly."

lib/xlsx/xlsx.js
contra: I believe this was just from running the project linter, if not I can revert this

contra:
@Siemienik Most of these changes are not going to be easy to land in a node 8 compatible way so I would suggest releasing this as a major or a part of the next major if one is planned.

lib/stream/xlsx/workbook-reader.js
```diff
@@ -76,16 +76,60 @@ class WorkbookReader extends EventEmitter {
     }
   }
 
+  async *flushQueue(force) {
+    if (!force && !this.isDirectReady()) return; // not ready yet! do nothing
+    if (this.waitingWorkSheets.length === 0) return; // queue flushed!
+    const queue = this.waitingWorkSheets; // make a copy
```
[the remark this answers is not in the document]
contra: @Siemienik `flushQueue` can be called multiple times asynchronously so there needs to be some mutable shared state with the worksheet queue on it - the alternative here would be passing an object in with the queue from the parent function. Based on everything else in this library it looked like mutable state living on the class was the preferred pattern.

lib/stream/xlsx/workbook-reader.js
contra: Addressed this above so going to mark this as resolved.

contra:
@Siemienik Rebased and addressed the PR feedback.

jcalderonzumba:
@yocontra @Siemienik is there any progress on this one?
We are happy users of your library and this PR will help us close an outstanding issue we have with some excel files that we process in our system.

Siemienik:
I would like too. However I am overloaded with my regular job.

yocontra:
@jcalderonzumba Could you test my fork against your workload and see if it resolves your issues? That would be helpful.

lucalas:
@Siemienik do you think there is a chance to review this pr and merge it?
We tried this fix and worked for us.

albanm:
I fully understand that sometimes time is lacking to maintain a library and that everybody here is working for free. But honestly these fixes are pretty important, IMO either these are fixed or the library should be considered as abandoned. At least there should be a disclaimer in the doc about the streaming parser having known issues, and the default options should be documented properly (styles being 'ignore' while documented as 'cache' is pretty bad).

Is there any hope of seeing this merged anytime ? Or does someone know a maintained fork that includes these fixes ?

yocontra:
@albanm You can always stick "github:yocontra/exceljs#fix-streams" in your package.json as the version to use this PR's version.

hcminh:
is there any update on this, I really need this fixed parser 😢

enjogole:
@yocontra I tried using your fork, but it still has a problem unless you opened and resaved the file, was this resolved?

Thanks.

mo-gad:
@Siemienik: This fix solves the issue we face as well (verified with "github:yocontra/exceljs#fix-streams"). Would really appreciate a final review and merge. Thanks a lot!

moonjoungyoung:
Hello. Any changes on this PR?

yocontra:
Would one of the maintainers want to take over this PR and get it across the finish line? It seems like it needs some very minor love. Due to a new role I'm not able to allocate any time to this - but would like to see it land eventually since we've been going back and forth on it (and me rebasing it) for a little over 4 years.

Totally understand if maintainers are busy and maybe somebody else wants to fork off my branch and resubmit instead.

matteovivona:
@yocontra I used `pnpm patch` to create a local patch with your PR

Issue #1305, opened 2020-05-27
[BUG] Errors when opening file in Excel after saving a file with conditional formatting


glen-84:
## 🐛 Bug Report

Lib version: 3.9.0

## Steps To Reproduce

```javascript
const Excel = require("exceljs");
const fs = require("fs");

const inputFilePath = "input.xlsx";
const outputFilePath = "output.xlsx";

(async () => {
    const workbook = new Excel.Workbook();

    const data = await fs.promises.readFile(inputFilePath);

    await workbook.xlsx.load(data.buffer);

    // Write.
    await fs.promises.writeFile(outputFilePath, await workbook.xlsx.writeBuffer());
})().catch((e) => console.error(e));
```

[input.xlsx](https://github.com/exceljs/exceljs/files/4689864/input.xlsx)
[output.xlsx](https://github.com/exceljs/exceljs/files/4689865/output.xlsx)

Opening the output file with Excel results in the following:

![image](https://user-images.githubusercontent.com/261509/83035404-b2b4e100-a039-11ea-9a1a-db08939a7f4c.png)

```xml
<? version="1.0" encoding="UTF-8" standalone="yes"?>
<recoveryLog xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
  <logFileName>
    error043120_01.xml
  </logFileName>
  <summary>
    Errors were detected in file 'C:\Users\Rob\Documents\glen-tmp\output.xlsx'
  </summary>
  <removedParts summary="Following is a list of removed parts:">
    <removedPart>
      Removed Part: /xl/styles.xml part with XML error.  (Styles) Load error. Line 2, column 6585.
    </removedPart>
  </removedParts>
  <repairedRecords summary="Following is a list of repairs:">
    <repairedRecord>
      Repaired Records: Cell information from /xl/worksheets/sheet2.xml part
    </repairedRecord>
    <repairedRecord>
      Repaired Records: Column information from /xl/worksheets/sheet2.xml part
    </repairedRecord>
    <repairedRecord>
      Repaired Records: Conditional formatting from /xl/worksheets/sheet2.xml part
    </repairedRecord>
  </repairedRecords>
</recoveryLog>
```

Just opening and saving the file causes the file size to go from 33 KB to 14 KB. 😮

If I clear the conditional formatting in the input file, then I don't see the errors (although the file size still drops dramatically, which is a bit worrying).

## The expected behaviour:

[...]

PR #1574, opened 2020-12-20
Fixed conditional format corrupting sheet #1305

rolandostar:
## Summary

I'm using a template for styling, and filling out data with ExcelJS. When my template had conditional formatting enabled on it, opening it and writing the file (even without data being inserted) would corrupt the file and lose styling. While investigating the issue I ran into [this comment](https://github.com/exceljs/exceljs/issues/1305#issuecomment-710115232) I tried it and it worked, and since I wasn't able to find a PR mentioning this fix I decided to open one myself.

## Test plan

My original template file's conditional formatting:

After being loaded by ExcelJS and rewritten into a different file:

image
image

## Notes

I understand the README.md mentioned writing tests for all fixes but I'm not really familiar with it, I can write them if someone could point me in the right direction.


mitchellwagner:
I can confirm that this fixes the issue with conditional styling and the corruption of the worksheet

Siemienik (approved):
I confirm that the new order is correct:
https://github.com/exceljs/ooxml-xsd/blob/aba7f40dc52d9a3cfe236e9e3e15a5148d5ca999/sml-styles.xsd#L830-L868

Thank you for that great contribution! 🥇 LGTM.

Issue #1575, opened 2020-12-28
[Issue] inlineStr is not supported


drdmitry:
## 🚀 Feature Proposal

Cell type `inlineStr` is not supported by exceljs.
Lib version: 4.2.0

## Motivation

If xlsx file has cells defined as `<c ... t="inlineStr"><is><t>value here</t></is></c>` - exceljs will put all values as 'null value'.

Quick reference: https://stackoverflow.com/a/4740690/4299666

At this moment I can't share the file. And idk which program/library produce this type of file.

## Expectation

exceljs should support cell type `inlineStr`.

## Example

Going to submit pull request.

gaurav-cointab:
This is still not fixed in 4.3.0.... 

Can you please reopen it and check. We are using the npm release version.

PR #1576, opened 2020-12-28
inlineStr cell type support #1575

drdmitry:
## Summary

Implementing inlineStr cell type fixing issue #1575

## Test plan

Tested stream reading of the xlsx file that has cells defined as `<c ... t="inlineStr"><is><t>value here</t></is></c>` - values gets populated, instead of null values before the fix.


alubbe:
Could you please add a new test with that test file to make sure we don't lose this functionality over time?

drdmitry:
@alubbe, added integration test with the sample file.

drdmitry:
@Siemienik, could you please review and merge this PR?

Siemienik:
Yes, I will if only I found free time.

alubbe (approved):

sveisvei:
When will this be released?

gaurav-cointab:
when will this be release, because right now in "4.3.0" also this fix is not present. We are getting the inlineStr cell as "NULL", please make a small release for this issue, if possible... Currently we are using the npm release version. Do we have to pull from github directly and then use the fix, or will you be making a release in npm also ?

adrian-morariu-clarisoft:
@Siemienik please release a new version with this fix

Siemienik:
@adrian-morariu-modus-create see #2351 ,

PR #1618, opened 2021-02-19
Types/workbook and worksheet reader

newbish:
## Summary

This will update the type definitions for the streaming Workbook and Worksheet readers so that the event can be accessed without having to cast the object as any or something else.

## Test plan

This change only addresses missing information in type definitions

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1980,10 +1980,14 @@ export namespace stream {
 
 		class WorkbookReader extends Workbook {
 			constructor(input: string | import('stream').Stream, options: Partial<WorkbookStreamReaderOptions>);
+			input: import('stream').Stream;
+			options: WorkbookStreamReaderOptions;
+			stream: import('stream').Stream;
+			styles: any; // StylesXform
 			read(): Promise<void>;
 			[Symbol.asyncIterator](): AsyncGenerator<WorksheetReader>;
 			parse(): AsyncIterator<any>;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1980,9 +1980,19 @@ export namespace stream {
 
 		class WorkbookReader extends Workbook {
 			constructor(input: string | import('stream').Stream, options: Partial<WorkbookStreamReaderOptions>);
+			input: import('stream').Stream;
+			options: WorkbookStreamReaderOptions;
+			stream: import('stream').Stream;
```
Siemienik, line 1985: I could not find this field in a workbook reader. Could I please add a permalink into the source code?

index.d.ts
```diff
@@ -1980,9 +1980,19 @@ export namespace stream {
...
+			styles: any; // StylesXform
```
Siemienik, line 1986: Please add a `// todo` here ;)

index.d.ts
```diff
@@ -1994,12 +2004,21 @@ export namespace stream {
 
 		class WorksheetReader {
 			constructor(options: WorksheetReaderOptions);
+			name: string;
+			state: string;
```
Siemienik, line 2008: I could not find this field in a workbook reader. Could I please add a permalink into the source code?

Siemienik (commented):
@newbish Thank you for this typings update 👍 

I have trouble with confirming two fields are really exist. If I could please you, show me where they were defined.

PR #1629, opened 2021-03-01
Add type definition for lastColumn property (fixes #1453)

### LINKED ISSUE #1453: Add type for property of worsheet.lastcolumn

masterzeus05:
## Summary

As detected as missing in #1453, the lastColumn property has its type defined in this PR.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1137,6 +1137,11 @@ export interface Worksheet {
 	readonly columnCount: number;
 
 	/**
+	 * Get the last column in a worksheet
+	 */
+	readonly lastColumn: Column;
+
+	/**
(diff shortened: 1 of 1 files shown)


Siemienik (approved):
LGTM:) I confirmed that the field exists in a sourcode

https://github.com/exceljs/exceljs/blob/ae2d2e5a63cd2432b228ef5dba8be7ab6a0f3abf/lib/doc/worksheet.js#L263-L265

PR #1664, opened 2021-04-13
FIXES #713 Remove 'use strict' strings brute force to enable module packaging

### LINKED ISSUE #713: Uncaught EvalError:  'unsafe-eval' is not an allowed source of script in the following Content

laurisvan:
## Summary

The current ExcelJS fails with CSP unsafe code evaluation when exceljs is bundled with recent WebPack versions. It appears that the root cause is regenerator-runtime resorting to 'eval':ing the source code whenever it encounters "use strict". This is probably because "use strict" shouldn't be necessary with ES modules (they are strict by default).

I tried various workarounds, such as replacing all "use strict" clauses when babelifying code, having similar replacements with browserify etc. I am almost sure the root cause relates to browserify, e.g. it somehow ends up adding the "use strict" clauses. In absence of a better solution, I thought that brute force removal of "use strict" strings was the only option left, so I added a Grunt task for that. I tried the more popular "grunt-replace" module for this, but unfortunately it failed to process some of the libraries included (because of @@ character somewhere within the code).

I think the long term solution would be to replace the current build system with some other than browserify that would supply the needed polyfills, but I believe such an architectural change would be much bigger undertaking.

## Test plan

Steps to repeat:
1) Include the module in a Webpack build, insert `new CspHtmlWebpackPlugin(cspConfigPolicy, cspOptions)` into plugins array
2) `npm link` the exceljs module into your project
3) Build and load the web page

N/A

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -3,6 +3,7 @@
 module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-babel');
   grunt.loadNpmTasks('grunt-browserify');
+  grunt.loadNpmTasks('grunt-string-replace');
   grunt.loadNpmTasks('grunt-terser');
   grunt.loadNpmTasks('grunt-contrib-jasmine');
   grunt.loadNpmTasks('grunt-contrib-copy');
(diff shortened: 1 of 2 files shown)


laurisvan:
Note: we have been running this in production since submitting this MR. I would respect if we could get it into mainline, or at least a suggestion of an alternative for how to fix this blocking issue.

dmathisen:
It's not clear, but this could be related to #713 

It's still an issue, but there are a few suggested fixes (see [here](https://github.com/exceljs/exceljs/issues/713#issuecomment-873997340) and [here](https://github.com/exceljs/exceljs/issues/713#issuecomment-1581175494)). The second one (my comment) worked for me and, to me, seems to be the lowest impact fix.

Issue #894, opened 2019-07-18
Wrong image position for non-integer coordinates


r4tz52:
Hi all,
I'm trying to place 16x16 image at arbitrary place of the worksheet. I set col number to 1.5, but it looks like image appears at the beginning of "B" column PLUS half of default column width, not half "B" column width.

**Code:**
```
const exceljs = require('exceljs');

const workbook = new exceljs.Workbook();
const worksheet = workbook.addWorksheet();

worksheet.columns = [{ width: 10 }, { width: 100 }, { width: 50 }];

const imageData = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAA'
	+ 'AQCAYAAAAf8/9hAAADBklEQVR42n2Se0hTURzHv3f3Xh/VysdmzlXLaiaIZtnTtMh'
	+ 'qVmNUaNADXFlBDwqCHlD/VERRRmVJ6001wqCgl9ibXKYSZajhtNK9tG1OF97yzrbr'
	+ 'vbdb/ZU0D/w4cM6Pz5fzOT8Cg9ZinSFW2t5LNUEURT9E8S0IYTzHDey3VD1/MLifG'
	+ 'HyQt2iZmxd4FSGK6GUCYNkgaJpCbEwUBEGYW1drqQ0LyM5ZoCQI+ESRgLPjG1YVpC'
	+ 'NVq0BFZTM+NHqhVsn3vntbUxIWMGfO3HHebtb5vZ/HOGUkLM82Q64cCwgMitaZ8bL'
	+ 'OvsftbDkZ/gmqGepDxVM7V+eORMqUZOmWBu9nQGpGoaHJh5kGs5Hzf7oZFnArZ8TY'
	+ 'tY+uuxB0AF4vBvo8kkMO9HASPmcAV858zD1gsb0ZUqLYcFQMdVVD7GcgIymIkCEiJ'
	+ 'hI/bL14X2bV5r370TYkwHcwvVE5X5PBe+wQqFF/OmjSD9P5zhfbXrCLh/5GKnHq8d'
	+ '1rK/fqoxOZumuQ8wz6Qz8RnVqA6z4d9pku7Oux1p/4L0Cl1mwsLCy8cvaMJFkU0P+'
	+ '5HvaOz0hUJCAu82+wftlKNDU3VXa6bPp/AIoEla7YuOEpFUH/HhYcO3oY1X6gzAVs'
[...]

PR #1688, opened 2021-04-25
fix issue #894

SuyongSun:
## Summary
When I use this lib for excel exporting, I found that the issue with image exporting.
When add image with none integer coordinates, I get the wrong position in result file. I found others also facing this problem in issues like #894.

I make some test and lookup the data affter unziping the drawing.xml file. I found that the error is caused by the wrong colwidth and rowheight. The tl and br are set to float number, if the col and row are not resized, the result is ok. But when width or height of the cell, in which tl or br defined, were resized, the result were still calculated with the standard cell width and row. So we got the wrong result.

## Test plan
I read the code related with this problem. Finally, I found the reason occured in anchor.js. There is my fix below.

Put the following code in the first line in constructor function. After this we can get the real column width and row height in worksheet.
this.worksheet = worksheet;

Then, modify the colWidth and rowHeight getters methods to get the correct number. As the default height or width can be undefined in worksheet properties, so we can add some protection code for below sample. We can check if the default heigth or width defined, if not, use constant value instead. (for example, height is 15, width is 9.14XXXXXX.)
Math.floor(this.worksheet.getRow(this.nativeRow + 1).height/this.worksheet.properties.defaultRowHeight * 180000) Math.floor(this.worksheet.getColumn(this.nativeCol + 1).width/this.worksheet.properties.defaultColWidth * 640000)

as the following code we put the image in xlsx with correct positions.
worksheet.addImage(imageId, { tl: {col: 1.1, row: 5.83 }, br: {col: 8.85, row: 10.2 } });
testimage
[https://user-images.githubusercontent.com/12286518/111744731-a64e8d00-88c6-11eb-827f-bff3231177ba.png](url)

### DIFF
--- a/lib/doc/anchor.js
+++ b/lib/doc/anchor.js
@@ -4,6 +4,7 @@ const colCache = require('../utils/col-cache');
 
 class Anchor {
   constructor(worksheet, address, offset = 0) {
+    this.worksheet = worksheet;
     if (!address) {
       this.nativeCol = 0;
       this.nativeColOff = 0;
@@ -29,8 +30,6 @@ class Anchor {
       this.nativeRow = 0;
       this.nativeRowOff = 0;
     }
-
-    this.worksheet = worksheet;
   }
 
   static asInstance(model) {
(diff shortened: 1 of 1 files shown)


SuyongSun:
The master code can not pass npm test successfully with node 8.17.0 on ubuntu. So it seems that not mine modification causing this problem.
And, I got the successful result with node 12.22.1 on local.

My local os is ubuntu 20.04.2 LTS

PR #1700, opened 2021-05-05
(docs): set prototype of RegExp correctly.

joeldenning:
## Summary

Today myself and others diagnosed an issue related to [the documentation for IE11]( https://github.com/exceljs/exceljs/blob/ae2d2e5a63cd2432b228ef5dba8be7ab6a0f3abf/README.md#es5-imports).

The existing documentation is incorrect because it changes the RegExp function such that `RegExp.prototype.exec` is no longer defined. After the exceljs IE11 polyfill, one must use `RegExp.prototype.prototype.exec` instead of `RegExp.prototype.exec` when trying to call exec directly. The reason is that `RegExp.prototype` is set to `RegExp` instead of `RegExp.prototype`, which adds an extra layer to the prototype chain.

It is reasonable to expect `RegExp.prototype.exec` to be defined, and for the API for using regular expressions to be unchanged after applying the polyfill in the exceljs documentation.

The fix here restores normal behavior for RegExp.prototype.exec.

## Test plan

We verified this in our production application. But since this a bug in the documentation, there are no automated tests included in my PR.

N/A


Siemienik (approved):

Siemienik:
LGTM, Thank you @joeldenning

Issue #1594, opened 2021-01-22
Stream xlsx return a corrupted file [Excel js 4.2.0]


Djcharles26:
I have a code in which i create a pretty big xlsx, for that I needed to use stream function, avoiding heap memory errors, everything worked fine on local environment, but in the moment I passed to a server, the stream files always returns a corrupted file, which I can only open in Excel with the repair tool, and that doesn't work every time.
At the same time, if the file is created with xlsx.writeFile(); it can be opened correctly. 
My server specs are:
### Code: 
```javascript
        let min = new Date(date.min);
        let max = new Date(date.max);
        const dir = `/home/data/uploads/reports/${req.user._id}/`;
        if(!fs.existsSync(dir))shelljs.mkdir("-p", dir);
        const filename = `report-${(new Date().getTime())}.xlsx`;
        const file = `${dir}/${filename}`;
        const stream = fs.createWriteStream(file);
        const options = {
            stream: stream,
            useStyles: true,
        };

        const workbook = new Exceljs.stream.xlsx.WorkbookWriter(options);
        const sheet = workbook.addWorksheet("reports");
       // process here (committing every row I create)
       // then committing the unique sheet I need
       sheet.commit();
       await workbook.commit();
```
Please help

m-janos:
I got similar issue. Rolling back the version of Node.js (from 15.6.0) to some older helped me. The exact cause is still unknown.

Djcharles26:
> I got similar issue. Rolling back the version of Node.js (from 15.6.0) to some older helped me. The exact cause is still unknown.

With what version of node did you achieve to work, I rolled back to 15.0 but didn't work neither

m-janos:
> > I got similar issue. Rolling back the version of Node.js (from 15.6.0) to some older helped me. The exact cause is still unknown.
> 
> With what version of node did you achieve to work, I rolled back to 15.0 but didn't work neither

13.10 It has to be under 15.

trypton:
Same here with nodejs 15.7, downgrading to nodejs 14 helped. The problem is that all files (styles, theme, workbook) in the xlsx container have 0 bytes size (except `sheet1.xml`).

germantellezv:
Same here with nodejs 10

alexbjorlig:
Did anyone find a work-around to avoid downgrading node.js?

ybonnefond:
[...]

PR #1731, opened 2021-05-31
Added Node v16 to the test suite

alexbjorlig:
## Summary

v16 is a new major version of Node.js, let's make sure to add test case of this.

Related to (at least) issue #1594

### DIFF
--- a/.github/workflows/exceljs.yml
+++ b/.github/workflows/exceljs.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         # https://github.com/actions/setup-node/issues/27
-        node-version: [8.17.0, 10.x, 12.x, 14.x]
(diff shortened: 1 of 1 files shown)


Siemienik (approved):
Thank you

alexbjorlig:
Should I add some timeouts to the Github actions, so they don't run for more than x minutes?

Siemienik:
> Should I add some timeouts to the Github actions, so they don't run for more than x minutes?

Yeah, it a good idea. Test runner sometimes gets stuck. I have no idea why, so the timeout will work fine.
However, it out of the scope of this PR. I am going to merge it as is. Could you set the timeout in another PR?

alexbjorlig:
@Siemienik  I created one here #1733. When this PR is merged, could you provide me with some basic help to get a failing test up and running? Because the simple writable stream example from the Read.me does not work with Node.js v16 🧐

PR #1733, opened 2021-06-01
Added timeouts to github actions

alexbjorlig:
## Summary
No need for actions running hour after hour - doing nothing  🌎  ☘️


cclauss (approved):

Siemienik:
There is no problem with a never-ending timeout, so I don't see the value of these changes. Could you explain your motivation?

cclauss:
Global warming?!?  Why cook the planet with tests that do not deliver value?

Siemienik (approved):
We have never situation like that. But ok, for the same reason, there are no concerns to introduce timeouts.

If are you really care about global warming, feel invited to open a PR with performance optimization. This will affect for much more apps than only a few actions.

Issue #1153, opened 2020-03-03
[Q] How to remove cell comments


Komojo7:
I've added a comment to a cell as follows:

```
ws.getCell('A1').note = {
    texts: [
        {
            text: 'some comment',
        },
    ],
};
```

I'm then trying to completely remove the comment afterwards, but i couldn't see any obvious way to do this.

Siemienik:
check this: https://github.com/exceljs/exceljs#cell-comments

Siemienik:
ws.getCell('A1').note = {};

should works;

dolbyzerr:
Seems that `ws.getCell('A1').note = {}` doesn't remove the note. It is still there but just empty, which results in having an empty note in the xlsx file.

The only workaround I found so far is
```
ws.getCell('A1')._comment = undefined
```

PR #1746, opened 2021-06-12
Add removeNote method exceljs#1153

dolbyzerr:
## Summary
There is no way to remove a note currently. I found related question: https://github.com/exceljs/exceljs/issues/1153 

According to @Siemienik's comment
```
ws.getCell('A1').note = {};
```
should work. But it just sets the empty comment and not removing it, which can be seen in excel.

Screenshot 2021-06-12 at 20 26 55

## Test plan

I've added some unit tests to test that `_comment` property is removed from the Cell class.
I've added a `test-comment-remove.js` script that gets `comments.xlsx` file with comments and generates `test.xlsx` file without comments to check that it's working properly.

Screenshot 2021-06-12 at 20 29 10

### DIFF
--- a/lib/doc/cell.js
+++ b/lib/doc/cell.js
@@ -229,6 +229,10 @@ class Cell {
     this._comment = new Note(note);
   }
 
+  removeNote() {
+    delete this._comment;
+  }
+
   get text() {
     return this._value.toString();
   }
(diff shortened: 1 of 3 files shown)


elefevre:
Any idea when this will be released?
In the meantime, I'm using this ugly hack:
```
delete (cell.model as any).comment;
delete (cell as any)._comment;
```

Issue #1412, opened 2020-07-28
Tables in Streams?


MatthewPinheiro:
Hello,

I didn't find record of this anywhere in the documentation or looking through some of the issues here, but apologies if someone has already asked about this.

Is it expected that Worksheets created from a Workbook Stream don't have an `.addTable` function? I'm using version 4.1.1. If this is expected, are there any plans to add that functionality? I'm trying to create a scalable process by utilizing streams for its memory advantages, and imo tables are a _huge_ part of Excel's usefulness for easy data manipulation, so I was surprised to see this isn't supported.

For reference, I'm initializing the worksheet I'd add the table to with the following code:
```
    var workbook = new Excel.stream.xlsx.WorkbookWriter();
    var worksheet = workbook.addWorksheet();

    worksheet.addTable // undefined
```

pconnell99:
I second this.  There is no support for tables in the streaming system, but I need to be able to build tables in a streaming fashion?  I don't know enough about the format of the files to know if this is feasible.  Is there another way we can achieve this?  If addTable() didn't overwrite existing content but allowed it to be applied to a range, we could reload and apply the table over the top of the streamed rows.

paeolo:
Any update on this? This feature is really valuable and it would great to have it in exceljs. At least the docs should mention that's it's not present in the streaming version of exceljs.

YukiGibson:
I wasted quite a lot of time trying to debug why this was failing for me. In TS, the signature of `Worksheet` was still showing the `addTable()` method. Obviously I got:
```
TypeError: worksheet.addTable is not a function
```
An unpleasant surprise to be honest when I searched any issue about it.

alvaro-escalante:
I have the same problem, I am facing 502 from the server in production trying to write heavy excel files, and wan to use `Excel.stream.xlsx`, but I am also using `worksheet.addTable()` and it fails.

mrawdon:
@alvaro-escalante #1885 should address this issue if it gets merged in

atignola:
Is there a scheduled release to include the fix from @mrawdon for this issue? The latest available version was released in August of last year.

PR #1885, opened 2021-11-01
Add table support for streaming Mode Fixes issue #1412

mrawdon:
## Summary

Excel tables were not previously supported in streaming mode. This adds support for tables when using streaming mode.  This fixes the error from #1412 where the addTable method was undefined. 

## Test plan

Added a streaming table test in workbook-xlsx-writer.spec.js I exported the table testing functions(addTable,checktable) from worksheet-table.spec.js

### DIFF
--- a/lib/doc/table.js
+++ b/lib/doc/table.js
@@ -113,7 +113,7 @@ class Table {
 
   get height() {
     // height of the table data
-    return this.table.rows.length;
+    return this.table.rows.length + this.rowsCommitted;
   }
 
   get filterHeight() {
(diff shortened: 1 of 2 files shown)


lib/stream/xlsx/worksheet-writer.js
```diff
@@ -246,6 +252,7 @@ class WorksheetWriter {
     this._writeBackground();
     this._writeHeaderFooter();
     this._writeRowBreaks();
+    this._writeTables();
```
ctaschereau, line 255: Related to https://github.com/exceljs/exceljs/pull/2079 , the line this._writeTables(); needs to be after the legacyData otherwise the notes in a streamed table make the resulting file invalid in Excel

ctaschereau (changes requested):

PR #1901, opened 2021-11-10
Fix worksheet reader type definitions

SpudNyk:
Update to match implementation and allow usage of id, name
and state properties.

## Summary

Allows accessing the worksheet id name and state when using the stream reading in Typescript

## Test plan

N/A

Matches implementation:
https://github.com/exceljs/exceljs/blob/0f6d7657e71f1de3c1fc2a2a75776b7766f0576a/lib/stream/xlsx/worksheet-reader.js#L12-L49

Properties are assigned by workbook reader here:
https://github.com/exceljs/exceljs/blob/0f6d7657e71f1de3c1fc2a2a75776b7766f0576a/lib/stream/xlsx/workbook-reader.js#L292-L314

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2000,17 +2000,21 @@ export namespace stream {
 		interface WorksheetReaderOptions {
 			workbook: Workbook;
 			id: number;
-			entry: import('stream').Stream;
+			iterator: AsyncIterator<any>;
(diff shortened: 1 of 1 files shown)


Siemienik (approved):
LGTM

PR #1907, opened 2021-11-17
FIX assignStyle on Table.store

o100ja:
The current implementation sets any style properties directly to the cell instead of merging into the existing cell style.

The result is that if you set a style in a column of a Table model it will break the resulting `xlsx`

## Summary

I need to be able to set the `numFmt` style on a Worksheet table. If I do it direct as in this example:
```javascript
  ws.addTable({
    name     : "FlexReport",
    ref      : "A1",
    headerRow: true,
    totalsRow: true,
    style    : {
      theme         : "TableStyleMedium9",
      showRowStripes: true,
    },
    columns  : [{
      key         : "id",
      name        : "ID",
      filterButton: true,
    }, {
      key         : "name",
      name        : "Name",
      filterButton: true,
    }, {
      key         : "value",
      name        : "value",
      filterButton: true,
      style       : {numFmt: "#,#00.00"},
    }],
    rows,
  });
```
the exported `xlsx` file is broke with the following error: 
image

## Test plan

After reviewing the source code I've tried to do the same export with the following change, which produced the correct format:
```javascript
  ws.addTable({
    name     : "FlexReport",
    ref      : "A1",
    headerRow: true,
    totalsRow: true,
    style    : {
      theme         : "TableStyleMedium9",
      showRowStripes: true,
    },
    columns  : [{
      key         : "id",
      name        : "ID",
      filterButton: true,
    }, {
      key         : "name",
      name        : "Name",
      filterButton: true,
    }, {
      key         : "value",
      name        : "value",
      filterButton: true,
      style       : {style:{numFmt: "#,#00.00"}}, // <-- Note the change which works but will overwrite any existing cell.style
    }],
    rows,
  });
```

- https://github.com/o100ja/exceljs/blob/5b41d7ac608ceaf8b0f8df389d8c7398ee68fdbd/lib/doc/table.js#L186

### DIFF
--- a/lib/doc/table.js
+++ b/lib/doc/table.js
@@ -183,7 +183,7 @@ class Table {
     const assignStyle = (cell, style) => {
       if (style) {
         Object.keys(style).forEach(key => {
-          cell[key] = style[key];
+          cell.style[key] = style[key];
         });
       }
     };


Siemienik (commented):
Great job and fixation 👍  Thank you.

Could you please add at least one test to confirm that the bug will not appear in the future?

o100ja:
> Great job and fixation +1 Thank you.
> 
> Could you please add at least one test to confirm that the bug will not appear in the future?

Unfortunately I simply do not have the time to do so, especially considering that you already have somewhat of a test case:
https://github.com/exceljs/exceljs/blob/d786c4db66f0e06587a72d1fac0bead394542957/test/test-table.js#L54

The problem is that there isn't enough deep assertion on the finished `xlsx` file and that is not a trivial taks

jbuck-lineleap:
@Siemienik any chance of getting this merged? This is massively blocking me on a feature and i don't want to have to fork for this

jbuck-lineleap:
@Siemienik I've created an additional PR here https://github.com/exceljs/exceljs/pull/2649 with the same change, an update to the type declarations, and a test.  Could you please merge this ASAP as it is blocking major functionality for me

PR #1934, opened 2021-12-18
Fix read this.worksheet before assign it

ZyqGitHub1:
The anchor.js assign `this.worksheet = worksheet` at end of constructor, but the getter (such as rowHeight) used in constructor will read this.worksheet. so it will cause rowHeight fall back to default value (180000);
 
## Summary
When I want to add a image in a cell over a range, and I adjust the height of this row before. The nativeRowOff will not calculate use the row height.
```javascript
const Excel = require('exceljs');
const fs = require('fs').promises;

(async () => {
    await workbook.xlsx.readFile('./test.xlsx');
    const worksheet = workbook.getWorksheet(1);
    const row = worksheet.getRow(1);
    row.height = 400;
    const col = worksheet.getColumn('A');
    col.width = 200
    const image = await fs.readFile('./image.png')
    const imageId = workbook.addImage({
        buffer: image,
        extension: 'png',
    });
    worksheet.addImage(imageId, {
        tl: { col: 0.5, row: 0.5 },
        br: { col: 1, row: 1 }
    });
    await workbook.xlsx.writeFile('./result.xlsx');
})()
```
result is
图片

## Test plan
After I change anchor.js,
```javascript
const Excel = require('exceljs');
const fs = require('fs').promises;

(async () => {
    await workbook.xlsx.readFile('./test.xlsx');
    const worksheet = workbook.getWorksheet(1);
    const row = worksheet.getRow(1);
    row.height = 400;
    const col = worksheet.getColumn('A');
    col.width = 200
    const image = await fs.readFile('./image.png')
    const imageId = workbook.addImage({
        buffer: image,
        extension: 'png',
    });
    worksheet.addImage(imageId, {
        tl: { col: 0.5, row: 0.5 },
        br: { col: 1, row: 1 }
    });
    await workbook.xlsx.writeFile('./result.xlsx');
})()
```
the result is
图片
The image top-left will start from 0.5*row height

https://github.com/ZyqGitHub1/exceljs/blob/458f372609589e6b3665a5c955998b677df2037e/lib/doc/anchor.js#L7

### DIFF
--- a/lib/doc/anchor.js
+++ b/lib/doc/anchor.js
@@ -4,6 +4,8 @@ const colCache = require('../utils/col-cache');
 
 class Anchor {
   constructor(worksheet, address, offset = 0) {
+    this.worksheet = worksheet;
+
     if (!address) {
       this.nativeCol = 0;
       this.nativeColOff = 0;
@@ -29,8 +31,6 @@ class Anchor {
       this.nativeRow = 0;
       this.nativeRowOff = 0;
     }
-
-    this.worksheet = worksheet;
   }
 
   static asInstance(model) {


hfhchan-plb:
This change means that all of the existing images are now off because the actual row height and col width are now being used. The previous behaviour was arguably a bug, but this is a breaking change causing every single image to be positioned differently.

hfhchan-plb:
There really should be an API which allows us to do actual absolute positioning, i.e. fixed pixel offset from the cell or fixed pixel offset from top left of spreadsheet, instead of the current absolute positioning which is more like relative.

PR #1938, opened 2021-12-20
Fix loadin of table rows and ref from xlsx file

antoineacy:
## Summary

Only structure (columns) of tables are loaded when reading a xlsx file. The data is missing : rows and ref properties remain undefined.

## Test plan

A test file is commited. It tests that table data is properly loaded.


zurmokeeper:
@antoineacy Could you please move [over here](https://github.com/zurmokeeper/excelize)? I'd be happy to merge this feature.

PR #2001, opened 2022-03-11
fix xlsx parser can not compatible with rich text tags that are not closed

WingMrL:
## Summary
Fix xlsx parser can not compatible with rich text tags that are not closed.

## Test plan
Unable to read an xlsx containing a hyperlink described in rich text. Here's the error infos:
```
error:  TypeError: Cannot create property 'richText' on string ''
    at SharedStringXform.parseClose (/Users/wingmrl/Code/exceljs-demo/node_modules/exceljs/lib/xlsx/xform/strings/shared-string-xform.js:81:40)
    at SharedStringsXform.parseClose (/Users/wingmrl/Code/exceljs-demo/node_modules/exceljs/lib/xlsx/xform/strings/shared-strings-xform.js:111:24)
    at SharedStringsXform.parse (/Users/wingmrl/Code/exceljs-demo/node_modules/exceljs/lib/xlsx/xform/base-xform.js:67:21)
```

```javascript
'use strict';

/**
 * Minimum reproducible program
 */

const fs = require('fs');
const stream = require('stream');

const Excel = require('../excel');

const filename = process.argv[2];
const buffer = fs.readFileSync(filename);

const streamReadable = new stream.Readable();
streamReadable.push(buffer);
streamReadable.push(null);

workbook.xlsx
  .read(streamReadable)
  .then(allWorksheet => {
    console.log('allWorksheet: ', allWorksheet);
  })
  .catch(error => {
    console.error('something went wrong', error.stack);
  });

```


zurmokeeper:
@WingMrL  Could you please move [over here](https://github.com/zurmokeeper/excelize)? I'd be happy to merge this feature.

PR #2009, opened 2022-03-15
Fix addThemes

bno1:
## Summary

This function would not insert the default theme if model.themes is an empty object. This causes errors when opening the spreadsheet in Excel.


zurmokeeper:
@bno1  Could you please move [over here](https://github.com/zurmokeeper/excelize)? I'd be happy to merge this feature.

Issue #1130, opened 2020-02-14
How read a csv file take the first line as column header


Diluka:
Lib ver: 3.5.0

example file:
```csv
header1,header2
aaaa,bbbb
```

How read a csv file take the first line as column header

And make `row.getCell('header1').value` working

Siemienik:
for similar purposes I implemented `xlsx-import` for you should work: (this is typescript, but it's works as well with js)

1. install package
`npm i xlsx-import --save`

2. type definition
```typescript

interface MyData{
    Header1: string;
    Header2: string;
}

```

3. import csv and data
```typescript
const workbook = new Excel.Workbook();
const ws = await workbook.csv.readFile(filename);
ws.name = "csv";

const importer = new Importer(workbook);
const data= importer.GetAllItems<MyData>( {
  type: 'list',
  worksheet: 'csv',
  columns: [
    { index: 1,  key: 'Header1' },
    { index: 2,  key: 'Header2' },
  ],
   rowOffset: 1, //offset header row
 }); 

console.log(data[0].header1);
```

Diluka:
It's still depending on the `index`. If removed one or more rows in xlsx file, the program will not work.

Siemienik:
@Diluka could I ask you for more explanation? I don't understand what is the problem.

Diluka:
@Siemienik sorry, i mean no matter the index of `header1` is, i want to get its value by name

for example i want to both file works with the same code
```csv
header1,header2
aaaa,bbbb
```
```csv
header2,header1
bbbb,aaaa
```

PR #2080, opened 2022-06-17
fix issue #1695, #1130 reading csv files with headers

### LINKED ISSUE #1695: [BUG] throws error when map is provided in CSV reader and headers = true.
## 🐛 Bug Report

Error String : data.map is not a function

## Steps To Reproduce
Use any simple CSV for reading and provide following in options.
parserOptions.headers = true,
map =  the given map function in documentation 

Call csv.readFile on this.

```
const anotherBook = new excel.Workbook();

await anotherBook.csv.readFile(csvPath, {
    parserOptions: { headers: true },
    map: function (value, index) {
        switch (index) {
            case 0:
                // column 1 is string
                return value;
            case 1:
                // column 2 is a date
                return new Date(value);
            case 2:
                // column 3 is JSON of a formula value
                return JSON.parse(value);
            default:
                // the rest are numbers
                return parseFloat(value);
        }
    }
});
```

## The expected behaviour:

Either feature should be supported or error should be caught and described in better way.

Either feature should be supported or error should be caught and described in better way.

```javascript

```

consatan:
## Summary

Fixed #1695, #1130, reading csv files with `parserOptions.headers` option not `false`, will be throw `data.map is not a function` error. Because the data is an `object`.

Add `headers` envent listener on `fast-csv` to get the header columns, then set to `Worksheet.columns`

## Test plan

See the integration test

### DIFF
--- a/lib/csv/csv.js
+++ b/lib/csv/csv.js
@@ -84,8 +84,20 @@ class CSV {
 
       const csvStream = fastCsv
         .parse(options.parserOptions)
+        .on('headers', data => {
+          const columns = [];
+          data.map(key => columns.push({key, header: key}));
+          worksheet.columns = columns;
+        })
         .on('data', data => {
-          worksheet.addRow(data.map(map));
+          if (Array.isArray(data)) {
+            worksheet.addRow(data.map(map));
+          } else {
+            Object.keys(data).forEach(key => {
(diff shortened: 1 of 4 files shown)


dhruvv90:
@consatan thanks for the PR. This is being reviewed and I'll get back soon.

PR #2089, opened 2022-06-28
Fix #1959 Corrupted file generating by reading and modifing an existing XLSX file with tables

### LINKED ISSUE #1959: Corrupted file generating by reading and modifing an existing XLSX file with tables [BUG] XYZ
please help me !!!! this code generate me a xlsx corrupted file but i don't now why:

public withTables = async (matrix: any[][], cols: number, rows: number) => {
        try {
            let workbook = new Excel.Workbook();
            workbook = await workbook.xlsx.readFile('./template/import.xlsx');
            const worksheet = workbook.getWorksheet('Foglio1');
            const table = worksheet.getTable('Ciao');
            this.reference = (table as any).table.tableRef;
            const refArray: Array<string> = this.reference.split(':');
            const begin: string = refArray[0];
            if (begin.length > 2) {
                this.rowNum = parseInt(begin[1] + begin[2]);
            } else {
                this.rowNum = parseInt(begin[1]);
            }
            const finale: Array<any> = [];
            worksheet.getRow(this.rowNum).eachCell({includeEmpty: true}, (cell, colNumber) => {
                if (cell.value === null) {
                    finale.push(cell.value);
                }
            });
            for (let i = 1; i < matrix.length + 1; i++) {
                worksheet.insertRow(this.rowNum + i, finale.concat(matrix[i - 1]));
                console.log(this.rowNum);
            }
            return await workbook.xlsx.writeFile('export.xlsx');
        } catch (e) {
            console.log(e);
        }
    }

MatthisClavijo:
## Summary
When reading and modifying an xlsx file containing a table, the generated xlsx file is corrupted.
This is due to the table not complying with the validation() process when parsing (reading) the xlsx file. 
This PR aims to comply with the validation() process.

#### *(Reading the xlsx file)*
I have written an integration test that loads an xlsx file containing 2 sheets with 2 tables each.
It then checks if the tables in the first sheet have correct references, modifies the reference of the first table and checks if the reference is correctly updated after commit.
No modification is done to the second sheet.
#### *(Generating the modified xlsx file)*
It then writes the new xlsx file.
#### *(Reading the generated xlsx file)*
It loads the newly written file and checks if the tables in the first sheet have correct references to determine if the file is corrupted.


theRealSheng (commented):
I just tried it and it didn't work for me. Still giving me the same error.

MatthisClavijo:
@theRealSheng If you are modifying or adding columns to a table, make sure to have unique column names otherwise Excel will report it as a "corrupted file". I'm currently working on another pull request that will add a check on column names.

If your issue comes from something else, feel free to contact me to discuss about it as I would be interested in making a more robust fix :)

lib/doc/worksheet.js
```diff
@@ -867,8 +867,25 @@ class Worksheet {
     this._media = value.media.map(medium => new Image(this, medium));
     this.sheetProtection = value.sheetProtection;
     this.tables = value.tables.reduce((tables, table) => {
-      const t = new Table();
+      const t = new Table(this, null);
+      table.ref = table.tableRef;
+      table.headerRow = true;
```
andytson, line 872: This is actually a corruption as well. The bug is in that Excel gives `headerRowCount="0"` when there is a header anyway, and the presence of `headerRowCount` is the difference in the xml of the table. The uncorrupt fix would be to test for presence of the attribute in https://github.com/exceljs/exceljs/blob/master/lib/xlsx/xform/table/table-xform.js#L69

Otherwise tables that are hiding the header will end up skipping the first line

lib/doc/worksheet.js
```diff
@@ -867,8 +867,25 @@ class Worksheet {
...
+      for (let i = 1; i < height; i++) {
+        const rowValues = [];
+        const row = this.getRow(tl.row + i);
+        for (let j = 0; j < width; j++) {
+          const cell = row.getCell(tl.col + j);
+          rowValues.push(cell.value);
+        }
+        t.addRow(rowValues);
+      }
+      t.commit();
```
andytson, line 888: It may be better to move this code to the table.js as a named constructor, so this mapping is in the same place as where all other table mapping code is.

PR #2090, opened 2022-07-04
Tables and media update functions when inserting or deleting rows/columns in a worksheet

MatthisClavijo:
## Summary
This PR aims at updating tables and media when a row or a column is inserted or removed in a worksheet. This PR also adds a check for table column name uniqueness. This PR depends on precedent PR #2089.

#### Table update:
- Reference
- Rows addition or deletion
- Columns addition or deletion

#### Medium update:
- Reference

## Test plan

***(Reading the xlsx file)***

I have written an integration test that loads an xlsx file containing 2 sheets with 2 tables each, 1 sheet with 3 images and 1 sheet with a chart.

<br><ins>Tables</ins>
It checks if the tables in the first sheet have correct references, adds and deletes rows and checks if the table's reference and rows are correctly updated.
It then adds and removes columns and checks if the table's reference and columns are correctly updated.

<br><ins>Media</ins>
It checks if the second sheet's media have correct references (anchor points), adds and deletes rows and checks if media's references are correctly updated.
It then adds and removes columns and checks if media's references are correctly updated.

No modification is done to the third and fourth sheet. Note that if you generate the output file, the chart doesn't appear as it is not yet supported in exceljs.


JoshuaBevers:
I was unable to resolve corrupted file errors with this fix.

MatthisClavijo:
Hi @JoshuaBevers, could you share with us your tests as I am looking forward to making a more robust fix if I missed something. Feel free to contact me by email if you want to discuss it in private :)

saliboubu:
Hi, I have been modifying .xml documents to be able to add charts on excel since it is not supported in exceljs and also found corruption problems when trying to open the excel. Is it possible to add charts now with the library? If so, where can I found the documentation?

Issue #1732, opened 2021-05-31
[BUG] Conditional Formatting is changing the font to italics


dancolq:
## 🐛 Bug Report

When loading and saving an excel file with conditional formatting, the font style is changing

From this 
![image](https://user-images.githubusercontent.com/16645222/120246785-32871400-c269-11eb-86bf-2522b9613ab1.png)
To this
![image](https://user-images.githubusercontent.com/16645222/120246845-63ffdf80-c269-11eb-89f2-44c528666e14.png)

Lib version: 4.2.1

## Steps To Reproduce

1) Add conditional formatting 
![image](https://user-images.githubusercontent.com/16645222/120247026-061fc780-c26a-11eb-877b-589cfc902a0c.png)
2) With these styles
![image](https://user-images.githubusercontent.com/16645222/120247057-23ed2c80-c26a-11eb-86e8-3e84c24cdcff.png)
3) Make sure the cell matches the criteria and save
4) Load the excel workbook with excel.js
5) Export the excel workbook
6) The style has changed to Bold Italic

## The expected behaviour:

The font style of the conditional formatted cell should be the same style as the original workbook prior to excels loading it.

## Possible solution :
I'm not sure, although I must say that I'm grateful for the previous conditional formatting bug to have been fixed in the first place.

sergeyampo:
I confirm that. If I set a rule style in conditional formatting, in debug mode I'll see that font style will be set like that:
worksheet.conditionalFormattings[i].rules[i].style.font.strike/bold/italic will be true but in the source excel file there will be a plain text.

Codeneos:
I can confirm this as well and have tracked back the root cause of this to the way boolean are parsed in styles.xml/dxf. The solution is straightforward so I'll try creating a fix and PR for this later.

In summary when you create a conditional formatting rule with a new style an entry under created in styles.xml dxfs tag that looks like this:
```xml
 <dxfs count="1">
        <dxf>
            <font>
                <b val="0"/>
                <i/>
                <strike val="0"/>
                <color theme="0" tint="-0.499984740745262"/>
            </font>
            <fill>
                <patternFill patternType="none">
                    <fgColor indexed="64"/>
                    <bgColor auto="1"/>
                </patternFill>
            </fill>
        </dxf>
    </dxfs>
```

[...]

PR #2095, opened 2022-07-15
fix: fix opening a worksheet with conditional formatting can cause bo…

Codeneos:
## Summary
When opening a workbook with conditional formatting rules and saving it back font style applied by when the conditional formatting rule is true can change. 

To reproduce the issue create workbook using Excel (latest version) with a conditional formatting rule that enables all font styles (italic, bold, strikethrough, etc). The condition of the rule is not relevant just make sure it applies to some rows in the sheet. Save the workbook, close it and reopen it in Excel. Now disable some of the font styles that are applied through the conditional formatting rule. Save the workbook and close it.

Now open the workbook using exceljs and save it without making any changes.

Expected behavior should be that the conditional formatting is exactly the same, observed behavior is that the cells to which conditional formatting is applied will have **all** font styles (italic, bold, strikethrough, etc) applied.

After analysis I found that the dfxs tag in styles.xml containing conditional styles was not getting parsed properly. Each font style is parsed as Boolean tag. When a font should be **bold** the `b` tag is present with no value or attributes: `<b/>`. When the bold is unchecked a `val` attribute is added and set to 0 making the tag look as follows: `<b val="0"/>`.

Looking at the font xform parser the configuration on first inspection looks correct, i.e:
```js
this.map = {
      b: {prop: 'bold', xform: new BooleanXform({tag: 'b', attr: 'val'})},
      i: {prop: 'italic', xform: new BooleanXform({tag: 'i', attr: 'val'})}
      ...
```

Looking a bit deeper in the `BooleanXform` class the `attr` property is being set but completely ignored during the actual parsing of the tag and always returns true. And that is correct when the value is actually true but incorrect when there is an `val` attribute on the tag. In which case `BooleanXform` should honor the value in `val` instead of just checking if the 

```js
class BooleanXform extends BaseXform {
  constructor(options) {
    super();

    this.tag = options.tag;
    this.attr = options.attr;
  }
  
  ...

  parseOpen(node) {
    if (node.name === this.tag) {
      this.model = true;
    }
  }
```

This also fixes issue #1732

## Test plan
We use exceljs to generate a workbook from a database using an excel file as template that defines which columns to export. The template excel file as conditional formattings that makes certain rows as italic. The formatting is copied from the template sheet to the new sheet. The proposed fix has been tested and is confirmed to work for our application.

The spec file that tests the Boolean parser is also updated to test `val='0'` is parsed as false.

The `BooleanXform`-class which is changed in the PR is only used in `FontXform` therefore I expect this change will not have any side effects. Apart from that these changes are backward compatible and only when there is an attribute will the attribute value be honored instead of just checking if the tag exists.

### DIFF
--- a/lib/xlsx/xform/simple/boolean-xform.js
+++ b/lib/xlsx/xform/simple/boolean-xform.js
@@ -17,7 +17,11 @@ class BooleanXform extends BaseXform {
 
   parseOpen(node) {
     if (node.name === this.tag) {
-      this.model = true;
+      if (this.attr) {
+        this.model = BaseXform.toBoolValue(node.attributes[this.attr], true);
+      } else {
+        this.model = true;
+      }
     }
   }
 
--- a/spec/unit/xlsx/xform/simple/boolean-xform.spec.js
+++ b/spec/unit/xlsx/xform/simple/boolean-xform.spec.js
@@ -16,13 +16,22 @@ const expectations = [
     tests: ['render', 'renderIn', 'parse'],
   },
   {
-    title: 'false',
+    title: 'zero',
     create() {
       return new BooleanXform({tag: 'boolean', attr: 'val'});
     },
     preparedModel: false,
-    xml: '',
-    tests: ['render', 'renderIn'],
+    xml: '<boolean val="0"/>',
+    tests: ['render', 'renderIn', 'parse'],
+  },
+  {
+    title: 'one',
+    create() {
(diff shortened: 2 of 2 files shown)


Codeneos:
This might also fix a few other issues related to font styles changing after opening a workbook with exceljs which include: `condensed`, `outline`, `shadow`, `extended`.

Dzorogh (approved):
Works! Thanks for patch

PR #2116, opened 2022-08-19
Add duplicate multiple rows feature

joaojhgs:
## Summary

The current duplicateRow functions does not work with multiple rows as it's supposed, it seems too functionally specific and allows little flexibility towards where to paste the duplicates.

This new implementation can duplicate single rows just like before, but it also allows to duplicate multiple rows as a group, and the new/generic parameters gives it more usage flexibility.

The user will be able to control which rows to duplicate, how much duplicates will be generated and where to paste them, the copies will be pasted in sequence, but the function could also be used multiple times pasting duplicates in different locations.

## Test plan

Initial Template:
image

Usage example:
`sheet.duplicateRows(
            {
                initialRow: 14,
                length: 4,
                firstRowToPaste: 17,
                copiesAmount: 2,
            },
        );`

Result:

image

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1239,7 +1239,12 @@ export interface Worksheet {
 	/**
 	 * Duplicate rows and insert new rows
 	 */
-	duplicateRow(rowNum: number, count: number, insert: boolean): void;
+	duplicateRows({
+		initialRow: number,
+		length = 1,
+		firstRowToPaste = 1,
+		copiesAmount = 1,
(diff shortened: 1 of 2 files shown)


joaojhgs:
My prettier ended up updating the worksheet file formatting, the only actual change is the removal of the duplicateRow func and the new duplicateRows func.

YangYongAn:
Looking forward to merging this feature

PR #2126, opened 2022-08-31
add characters cannot be used for worksheet name

tkm-kj:
## Summary

As for worksheet name, it raises error if worksheet name contains followings characters now

```js
'*', '?', ':', '/', '\\', '[', ']'
```

I expect this implementation prevents worksheet from breaking
But, if these characters are full-width characters, the worksheet is broken

https://user-images.githubusercontent.com/3395349/187575650-5528a378-755e-46ae-a3fa-d12925aee154.mov

So, I fixed the issue

## Test plan

Code example is below

```js
import ExcelJS from 'exceljs';

const workbook = new ExcelJS.Workbook()
workbook.addWorksheet('？')
```

If error raises, the test is success

### DIFF
--- a/lib/doc/workbook.js
+++ b/lib/doc/workbook.js
@@ -60,9 +60,9 @@ class Workbook {
 
     // Illegal character in worksheet name: asterisk (*), question mark (?),
     // colon (:), forward slash (/ \), or bracket ([])
-    if (/[*?:/\\[\]]/.test(name)) {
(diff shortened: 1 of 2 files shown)


coder-xiaotian (approved):

Siemienik (approved):
LGTM, thank you. I think we should check if else characters are incorrect. I suppose there is only a narrow range of allowed characters. Will create an investigation ticket for that. (https://github.com/exceljs/exceljs/issues/2243)

naorpeled (approved):

Issue #2184, opened 2022-12-26
[BUG] Excel shows error when reading generated file


nbelyh:
## 🐛 Bug Report

Excel does not open the file generated by exceljs. When opening, it shows the error below

![image](https://user-images.githubusercontent.com/528366/209547935-d02cf325-86f5-4b83-b72c-eee8dd52e03c.png)

Error log (the meaningful part of it):
```
<removedRecord>Removed Records: AutoFilter from /xl/tables/table1.xml part (Table)</removedRecord>
```

Lib version: 4.3.0

## Steps To Reproduce

Use an Excel file that has a **Table** and **AutoFilter** enabled for that table, and run the following code:

![image](https://user-images.githubusercontent.com/528366/209548246-afff1b85-90e1-48d4-b938-baa8babca736.png)

```javascript
  let workBook = new ExcelJS.Workbook()
  await workBook.xlsx.readFile("./x.xlsx");
  await workBook.xlsx.writeFile("./y.xlsx");
```

When opening "y.xlsx" the above error occurs.

## The expected behaviour:

The file is opened without any issues.

## Possible solution (optional, but very helpful):

Please see the **linked PR** that fixes the problem.

The issue is caused by invalid combination written into excel file. Original file is on the left, saved is on the right:
![image](https://user-images.githubusercontent.com/528366/209548633-1227630b-0b7d-4dff-815d-35cd8ee2e169.png)

The problem is library reaction on "undefined" (missing) value in the source file.
It uses wrong default ("1") instead of missing value. Another thing is it turns "0" into "1" for the header, but that does not block opening.

trylovetom:
I use your version https://github.com/exceljs/exceljs/pull/2185, but it still exists.
```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<recoveryLog xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
  <logFileName>Repair Result to 0.xml</logFileName>
  <summary>Errors were detected in file '/Users/tzu-yen/Downloads/「大象班」的「批閱」成績報表 (65).xlsx'</summary>
  <removedFeatures summary="Following is a list of removed features:">
    <removedFeature>Removed Feature: AutoFilter from /xl/tables/table3.xml part (Table)</removedFeature>
    <removedFeature>Removed Feature: Table from /xl/tables/table3.xml part (Table)</removedFeature>
  </removedFeatures>
</recoveryLog>
```

[...]

PR #2185, opened 2022-12-26
fix file not opening because of wrong defaults

nbelyh:
Fixes reading the default values #2184

## Summary

This fixes a bug with Excel showing an error when reading the generated file.

## Test plan

Please refer to the linked issue #2184

### DIFF
--- a/lib/xlsx/xform/table/filter-column-xform.js
+++ b/lib/xlsx/xform/table/filter-column-xform.js
@@ -4,6 +4,8 @@ const ListXform = require('../list-xform');
 const CustomFilterXform = require('./custom-filter-xform');
 const FilterXform = require('./filter-xform');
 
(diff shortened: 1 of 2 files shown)


lib/xlsx/xform/table/filter-column-xform.js
```diff
@@ -4,6 +4,8 @@ const ListXform = require('../list-xform');
 const CustomFilterXform = require('./custom-filter-xform');
 const FilterXform = require('./filter-xform');
 
+const undefinedOrResult = (test, result) => (typeof test === 'undefined' ? undefined : result);
```
tanvirstreame, line 7:
```suggestion
const result = (test, result) => (typeof test === 'undefined' ? undefined : result);
```
I think better keeping it simple to `result`

nbelyh: "result" is too generic and does not help to understand the purpose of the function?

PR #2244, opened 2023-04-14
fix: Fix xlsx.writeFile() not catching error when error occurs

zurmokeeper:
## Summary
When `await workbook.xlsx.writeFile(filename).`
was reported, but the error was not caught by try catch, after modifying the source code, it can now be caught

## Test plan
```
try {
    const imageId2 = workbook.addImage({
        buffer: fs.readFileSync('path/to.image.png'),
        extension: 'png',
      });
    await workbook.xlsx.writeFile(filename);
} catch (error) {
    console.log(error)
}
```

### DIFF
--- a/lib/xlsx/xlsx.js
+++ b/lib/xlsx/xlsx.js
@@ -675,6 +675,8 @@ class XLSX {
 
       this.write(stream, options).then(() => {
         stream.end();
+      }).catch((err)=>{
+        reject(err);
       });
     });
   }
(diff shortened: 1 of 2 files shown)


Siemienik (commented):
Good catch, looks good for me, but test case is required before merge

zurmokeeper:
> Good catch, looks good for me, but test case is required before merge

Great, I will add the test cases later

zurmokeeper:
@Siemienik Additional test cases have been added

naorpeled (approved):

Siemienik (approved):
LGTM, thanks

PR #2284, opened 2023-06-08
fix: add spinCount field for WorksheetProtection type

damingerdai:


index.d.ts
```diff
@@ -880,6 +880,7 @@ export interface WorksheetProtection {
 	sort: boolean;
 	autoFilter: boolean;
 	pivotTables: boolean;
+  spinCount: number;
```
Siemienik, line 883:
```suggestion
	spinCount: number;
```

Siemienik (approved):
LGTM

Issue #2303, opened 2023-07-12
[BUG] Out-of-memory with lots of cells


hfhchan-plb:
## 🐛 Bug Report

node crashes with cryptic error messages like:

```
#
# Fatal error in , line 0
# Fatal JavaScript invalid size error 169220804
#
#
#
#FailureMessage Object: 0x17ccb4f08
 1: 0x100acbc2c node::NodePlatform::GetStackTracePrinter()::$_3::__invoke() [/Users/user/Library/Application Support/fnm/node-versions/v18.14.2/installation/bin/node]
 2: 0x1019fdd24 V8_Fatal(char const*, ...) [/Users/user/Library/Application Support/fnm/node-versions/v18.14.2/installation/bin/node]
 3: 0x100d3f1e8 v8::internal::FactoryBase<v8::internal::Factory>::NewFixedArrayWithFiller(v8::internal::Handle<v8::internal::Map>, int, v8::internal::Handle<v8::internal::Oddball>, v8::internal::AllocationType) [/Users/user/Library/Application Support/fnm/node-versions/v18.14.2/installation/bin/node]
 4: 0x100ecb4d0 v8::internal::(anonymous namespace)::ElementsAccessorBase<v8::internal::(anonymous namespace)::FastPackedObjectElementsAccessor, v8::internal::(anonymous namespace)::ElementsKindTraits<(v8::internal::ElementsKind)2>>::GrowCapacity(v8::internal::Handle<v8::internal::JSObject>, unsigned int) [/Users/user/Library/Application Support/fnm/node-versions/v18.14.2/installation/bin/node]
 5: 0x1010cb2f0 v8::internal::Runtime_GrowArrayElements(int, unsigned long*, v8::internal::Isolate*) [/Users/user/Library/Application Support/fnm/node-versions/v18.14.2/installation/bin/node]
 6: 0x10142104c Builtins_CEntry_Return1_DontSaveFPRegs_ArgvOnStack_NoBuiltinExit [/Users/user/Library/Application Support/fnm/node-versions/v18.14.2/installation/bin/node]
```

Lib version: 4.3.0

## Possible solution:

It appears that commenting out the following code in styles-xform.js resolves the issue.

```
if (this.weakMap) {
  this.weakMap.set(model, styleId);
}
```

hfhchan-plb:
See also: https://github.com/exceljs/exceljs/issues/2041.

It seems weak maps cause more issues than they solve.

[...]

PR #2319, opened 2023-07-29
Increase resilience to generating large workbooks

hfhchan-plb:
## Summary

Updates the xml-stream.js to generate a single string instead of multiple strings for each attribute or record of attributes.

Closes: #2303

### DIFF
--- a/lib/utils/xml-stream.js
+++ b/lib/utils/xml-stream.js
@@ -12,19 +12,17 @@ const QUOTE = '"';
 const SPACE = ' ';
 
 function pushAttribute(xml, name, value) {
-  xml.push(SPACE);
-  xml.push(name);
-  xml.push(EQUALS_QUOTE);
-  xml.push(utils.xmlEncode(value.toString()));
-  xml.push(QUOTE);
+  xml.push(` ${name}="${utils.xmlEncode(value.toString())}"`);
 }
 function pushAttributes(xml, attributes) {
   if (attributes) {
+    const tmp = [];
     _.each(attributes, (value, name) => {
       if (value !== undefined) {
-        pushAttribute(xml, name, value);
+        pushAttribute(tmp, name, value);
       }
     });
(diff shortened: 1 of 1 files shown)


zurmokeeper:
@hfhchan-plb Can you please tell me how to add test cases for this.

hfhchan-plb:
I'm not sure if adding any test case is realistic for this, because after the patch I can now generate a sheet with 18527 rows x 272 columns (with some cells with some formatting). Before this patch, I could only generate the sheet with 18527 rows x 31 columns. While generating the sheet the node process went over 4GB RAM.

hfhchan-plb:
OK it seems the limit for GitHub Actions is 7 GB (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners), so it should be feasible to write a use case which generates workbooks of that size.

hfhchan-plb:
A brief simulation:

```
const rowCount = 18527;
const colCount = 272;

for (let i = 1; i <= rowCount; i++) {
  const row = [];
  for (let j = 1; j <= colCount; j++) {
    row.push(Math.random());
  }
  worksheet.addRow(row);

  if (i % 33 === 1) { // Make header every 33 rows
    for (let col = 1; col <= colCount; col += 1) {
      const cell = worksheet.getCell(i, col);
      cell.font = { color: { argb: "FFFFFFFF" }, bold: true };
      cell.fill = {
        type: "pattern",
        pattern: "solid",
        fgColor: { argb: "#FFD9D9D9" },
      };
    }
  } else {
    for (let col = 1; col <= colCount; col += 1) {
      const cell = worksheet.getCell(i, col);
      cell.numFmt = '0.00;[Red]-0.00;-'
    }
  }
}
```

hfhchan-plb:
I renamed my branch to `hfhchan-plb:fix-large-file-export` and GitHub automatically closed this PR. Please see the new PR #2320.

PR #2551, opened 2023-10-09
Add pivot table with limitations

mikez:
```js
worksheet.addPivotTable(configuration);
```

**Note:** Pivot table support is in its early stages with certain limitations, including:

- No support for reading xlsx documents with existing pivot tables (writing is supported).
- Pivot table configurations must consist of one "value" item and use the sum metric.
- Only one pivot table can be added for the entire document.

## Example usage

```js

const worksheet1 = workbook.addWorksheet('Sheet1');
worksheet1.addRows([
  ['A', 'B', 'C', 'D', 'E'],
  ['a1', 'b1', 'c1', 4, 5],
  ['a1', 'b2', 'c1', 4, 5],
  ['a2', 'b1', 'c2', 14, 24],
  ['a2', 'b2', 'c2', 24, 35],
  ['a3', 'b1', 'c3', 34, 45],
  ['a3', 'b2', 'c3', 44, 45],
]);

const worksheet2 = workbook.addWorksheet('Sheet2');
worksheet2.addPivotTable({
  // Source data: entire sheet range
  sourceSheet: worksheet1,
  // Pivot table fields: via header row in `worksheet1`
  rows: ['A', 'B'],
  columns: ['C'],
  values: ['E'], // Exactly 1 field
  metric: 'sum', // Metric: 'sum' only
});
```


lib/doc/worksheet.js
```diff
@@ -806,6 +809,142 @@ class Worksheet {
...
+    //
+    // into
+    //
+    //  [
+    //    { name: 'A', sharedItems: ['a1', 'a2', 'a3'] },
+    //    { name: 'B', sharedItems: ['b1', 'b2'] },
+    //    { name: 'C', sharedItems: ['c1', 'c2', 'c3'] },
+    //    { name: 'D', sharedItems: null },
+    //    { name: 'E', sharedItems: null },
+    //
```
mikez, line 925: forgot the ]

lib/doc/pivot-table.js
```diff
@@ -0,0 +1,132 @@
...
+    throw new Error('No pivot table rows specified.');
+  }
+
+  if (!model.columns.length) {
+    throw new Error('No pivot table columns specified.');
+  }
+
+  if (model.values.length !== 1) {
+    throw new Error('Exactly 1 value needs to be specified at this time.');
+  }
```
Siemienik, line 82: Would It be enabled to add partially configured pivot table without for instance rows set?

lib/doc/pivot-table.js
```diff
@@ -0,0 +1,132 @@
...
+  //
+  //  [
+  //    { name: 'A', sharedItems: ['a1', 'a2', 'a3'] },
+  //    { name: 'B', sharedItems: ['b1', 'b2'] },
+  //    { name: 'C', sharedItems: ['c1', 'c2', 'c3'] },
+  //    { name: 'D', sharedItems: null },
+  //    { name: 'E', sharedItems: null }
+  //  ]
+
+  const names = worksheet.getRow(1).values;
```
Siemienik, line 113: Does it always start from row 1?

lib/doc/worksheet.js
```diff
@@ -806,6 +809,17 @@ class Worksheet {
     return Object.values(this.tables);
   }
 
+  // =========================================================================
+  // Pivot Tables
+  addPivotTable(model) {
+    const pivotTable = makePivotTable(this, model);
+
+    this.pivotTables.push(pivotTable);
+    this.workbook.pivotTables.push(pivotTable);
```
Siemienik, line 818: What will happen if someone remove the pivot table from one of these places, or overwrite one of them?
What do you think about making `wb.workbook.pivotTables` a   getter returning pivottables from all worksheets?  Or .. a function?
Can it attach one `pivotTable` to multiple sheets?

lib/doc/pivot-table.js
```diff
@@ -0,0 +1,132 @@
...
+
+  const aggregate = columnIndex => {
+    const columnValues = worksheet.getColumn(columnIndex).values.splice(2);
+    const columnValuesAsSet = new Set(columnValues);
+    return toSortedArray(columnValuesAsSet);
+  };
+
+  // make result
+  const result = [];
+  for (const columnIndex of range(1, names.length)) {
```
Siemienik, line 124: wdyt?
```suggestion
  for (const columnIndex of names.keys()) {
```

lib/doc/pivot-table.js
```diff
@@ -0,0 +1,132 @@
...
+  //    { name: 'E', sharedItems: null }
+  //  ]
+
+  const names = worksheet.getRow(1).values;
+  const nameToHasSharedItems = objectFromProps(fieldNamesWithSharedItems, true);
+
+  const aggregate = columnIndex => {
+    const columnValues = worksheet.getColumn(columnIndex).values.splice(2);
+    const columnValuesAsSet = new Set(columnValues);
+    return toSortedArray(columnValuesAsSet);
```
Siemienik, line 119: Has it to be sorted? What happen if  not?

lib/xlsx/xform/book/workbook-pivot-cache-xform.js
```diff
@@ -0,0 +1,29 @@
...
+
+  parseOpen(node) {
+    if (node.name === 'pivotCache') {
+      this.model = {
+        cacheId: node.attributes.cacheId,
+        rId: node.attributes['r:id'],
+      };
+      return true;
+    }
+    return false;
```
Siemienik, line 19: Not so important, only a little bit readability  improvement 😄 

```suggestion
    if (node.name !== 'pivotCache') {
       return false;
    }
    this.model = {
      cacheId: node.attributes.cacheId,
      rId: node.attributes['r:id'],
    };
    return true;
```

lib/xlsx/xform/pivot-table/pivot-table-xform.js
```diff
@@ -0,0 +1,191 @@
...
+  // fieldType: 'row', 'column', 'value', null
+
+  const defaultAttributes = 'compact="0" outline="0" showAll="0" defaultSubtotal="0"';
+
+  if (fieldType === 'row' || fieldType === 'column') {
+    const axis = fieldType === 'row' ? 'axisRow' : 'axisCol';
+    return `
+      <pivotField axis="${axis}" ${defaultAttributes}>
+        <items count="${sharedItems.length + 1}">
+          ${[...range(0, sharedItems.length)].map(index => `<item x="${index}" />`).join('\n              ')}
```
Siemienik, line 171: You may try this way:
```suggestion
          ${sharedItems.map((item, index) => `<item x="${index}" />`).join('\n              ')}
```

Siemienik (commented):
Hello @mikez , Thank you for the latest changes, I'm really happy that Exceljs has possibility to gain new feature even with limitations.
I added few comments, I would be glad for answering that 😄 

Good job and I love the direction when it comes. 🥇

lib/doc/pivot-table.js
mikez: At this time, the code doesn't support that. You need at least one item in each of {rows, columns, values}.
However, this could be a potential feature down the road if it is requested.

lib/doc/pivot-table.js
mikez: I made the simplifying assumption that the source data is represented by the entire sheet, with the first row representing the header. Do you this assumption is too simplistic and won't capture 95% of the cases?

lib/doc/worksheet.js
mikez: > What will happen if someone remove the pivot table from one of these places, or overwrite one of them?

@Siemienik It seems you're pointing at the mutability of `worksheet.pivotTables` and `worksheet.workbook.pivotTables` and concerns for that clients might mutate this. If you think this type of safety is needed and consistent with elsewhere in ExcelJS, we can definitely make it a getter. I'll take your advice here.

> Can it attach one pivotTable to multiple sheets?

At this time, I intend for there to be at most one pivotTable in at most one sheet across the entire workbook. The user should not have to mutate `worksheet.pivotTables` or `worksheet.workbook.pivotTables` and only access `worksheet.addPivotTable`.

lib/doc/pivot-table.js
mikez: This won't work, because the iterator will include the index 0 and we don't want that. (This is because of the design of `.getRow()` where we start counting at index 1 vs. index 0 which is customary otherwise in JavaScript.)

```js
> Array.from(worksheet1.getRow(1).values.keys())
[
  0, 1, 2, 3, 4,
  5, 6, 7, 8
]
```

lib/doc/pivot-table.js
mikez: It's sorted because the fields show up sorted in the pivot table. This makes the logic and indexing easier.

You can try this out by creating a column "A" with data a1, a2, a3, a4 scrambled. In the resulting pivot table, it will be sorted alphabetically.

lib/xlsx/xform/book/workbook-pivot-cache-xform.js
mikez: I see the pattern you points towards, and I agree in general! However, this is copy paste from elsewhere in ExcelJS. So I tried to stay consistent with @guyonroche's pattern of returning `true` first and `false` later. However, since pivot table doesn't handle "read" for now, this code won't get run anyways. :)

lib/xlsx/xform/pivot-table/pivot-table-xform.js
mikez: Clever. I like this. I will update and force push this.

mikez:
> @Siemienik: I added few comments, I would be glad for answering that 😄

Thank you. Please guide me where I might want to add clarifications to the documentation or comments to make it clearer for other readers.

lib/xlsx/xform/book/workbook-pivot-cache-xform.js
Siemienik: it's just a detail, nevermind

Siemienik (approved):
LGTM, thank you for booting up new feature into exceljs 👏

mikez:
👉 Feedback and discussion [here](https://github.com/exceljs/exceljs/discussions/2575).

PR #2558, opened 2023-10-17
Fix the problem with writing a file using streams and not using RAM

zlooun:
## Summary

 This PR fixes the problem with writing a file using streams and not using RAM. It solves the problem of writing large excel files.

## Test plan

```
import * as fs from 'fs';
import { stream } from 'exceljs';

const output_file_name = "/test.xlsx";

const writeStream = fs.createWriteStream(output_file_name, { flags: 'w' });
const wb = new stream.xlsx.WorkbookWriter({ stream: writeStream });
const worksheet = wb.addWorksheet("test");

const headers = Array.from({length: 256}, (_, i) => i + 1).map((i) => 'test' + i);

for (let i = 0; i < 100000; i++) {
  const row = headers.map((header) => header + '|' + i);
  await worksheet.addRow(row).commit(); // This raw will be immediately written to disk and will not clog RAM.
}

await worksheet.commit(); // This is not necessary because await wb.commit() is used, but you can also write to disk not raw by raw, but worksheet by worksheet.
await wb.commit();
```

https://github.com/zlooun/exceljs
https://www.npmjs.com/package/@zlooun/exceljs


Siemienik:
Thank you for the contribution; I believe it enhances the handling of large file writing 🎉. Please bear with me, as this change breaks backward-compatibility and will need to wait for the next major release (5.0).

AmberWilkie:
Hello just hoping to bump some attention here. This PR would be very helpful for my work. I'd offer to help get it in but honestly I learned everything I know about streams and buffers trying to understand why `await workbook.commit()` was hanging forever. In my case, it's because my S3 upload releases the stream, which I expected to do after committing the workbook. So effectively that `await` will be forever if I put it before the S3 upload terminates.

(Or then again, I could be off base here - like I said, new to streams).

julesblancphin:
Are there any plans to merge this PR?

PR #2562, opened 2023-10-19
fix: fixes typescript and intellisense

### LINKED ISSUE #2548: [BUG] Typescript typings don't work
## 🐛 Bug Report

Looks like typescript d.ts file is wrong: typings are not suggested at all (using webstorm). 

Lib version: 4.3.0

## Steps To Reproduce

```typescript
import { Workbook } from 'exceljs/index';
const workbook = new Workbook();
const worksheet = workbook.addWorksheet(name); // <--- unresolved function method "addWorksheet()"
```

It looks like *.ts file imports js implementation as "*" in index.ts file which is tricky and not well supported.

groozin:
## Summary

Fixes #2548 

Both intellisense and typescript tests were broken on `master`. Though the issue mentions webstorm as the editor I had the same with vscode. 

Converted the `index.ts` into `index.js` with CJS syntax as per `eslint` requirement. Also fixed the typescript tests and added them to `npm run test:full` script - now that they are working.

## Test plan

For testing I ran `npm run build`
```bash
 ~/r/exceljs | fix/2548/fix..intellisense                                                                                                                51s | 20.8.0 node | 2.7.5 rb
> npm run build

> exceljs@4.3.0 build
> grunt build

Running "babel:dist" (babel) task

Running "browserify:bare" (browserify) task
>> Bundle ./dist/exceljs.bare.js created.

Running "browserify:bundle" (browserify) task
>> Bundle ./dist/exceljs.js created.

Running "browserify:spec" (browserify) task
>> Bundle ./build/web/exceljs.spec.js created.

Running "terser:dist" (terser) task
>> 1 grunt.util.pluralize(createdFiles, 'file/files') created.

Running "terser:bare" (terser) task
>> 1 grunt.util.pluralize(createdFiles, 'file/files') created.

Running "exorcise:bundle" (exorcise) task
Exorcising source map from ./dist/exceljs.js
Exorcising source map from ./dist/exceljs.bare.js

Running "copy:dist" (copy) task
Created 20 directories, copied 332 files

Done.
```

Then `npm run test` (the output is really big so I omit it here).

Also to check intellisense working and to make sure that different module approaches are still working I created 3 files: one `ESM`, one `CJS` and one `TS`. I then use `npm install` from the local source code of my branch to have the changes in. And both checked the intellisens and ran the code in the files to see if the lib is working.

See below:
image

### DIFF
--- a/spec/typescript/exceljs.spec.ts
+++ b/spec/typescript/exceljs.spec.ts
@@ -1,7 +1,8 @@
 import "regenerator-runtime/runtime";
 
 import { expect } from "chai";
-import ExcelJS from "../../index";
+import * as ExcelJS from "../../index";
+import { createReadStream, createWriteStream } from "node:fs";
 
 describe("typescript", () => {
   it("can create and buffer xlsx", async () => {
@@ -18,23 +19,22 @@ describe("typescript", () => {
     const ws2 = wb2.getWorksheet("blort");
     expect(ws2.getCell("A1").value).to.equal(7);
   });
-  it("can create and stream xlsx", async () => {
+
+  it("can stream xlsx", async () => {
     const wb = new ExcelJS.Workbook();
     const ws = wb.addWorksheet("blort");
     ws.getCell("A1").value = 7;
 
-    const wb2 = new ExcelJS.Workbook();
-    const stream = wb2.xlsx.createInputStream();
-    await wb.xlsx.write(stream);
-    stream.end();
+    const write = createWriteStream("test.xlsx");
 
-    await new Promise((resolve, reject) => {
-      stream.on("done", () => {
-        const ws2 = wb2.getWorksheet("blort");
-        expect(ws2.getCell("A1").value).to.equal(7);
-        resolve();
-      });
-      stream.on("error", reject);
-    });
+    await wb.xlsx.write(write);
+
+    const read = createReadStream("test.xlsx");
+
+    const wb2 = new ExcelJS.Workbook();
+    await wb2.xlsx.read(read);
+    const ws2 = wb2.getWorksheet("blort");
+
+    expect(ws2.getCell("A1").value).to.equal(7);
   });
 });
--- a/spec/typescript/exceljs.spec.ts
+++ b/spec/typescript/exceljs.spec.ts
@@ -1,13 +1,13 @@
-import 'regenerator-runtime/runtime';
+import "regenerator-runtime/runtime";
 
-import { expect } from 'chai';
-import ExcelJS from '../../index';
+import { expect } from "chai";
+import ExcelJS from "../../index";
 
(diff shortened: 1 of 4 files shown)


spec/typescript/exceljs.spec.ts
```diff
@@ -1,40 +1,40 @@
-import 'regenerator-runtime/runtime';
+import "regenerator-runtime/runtime";
 
-import { expect } from 'chai';
-import ExcelJS from '../../index';
+import { expect } from "chai";
+import * as ExcelJS from "../../index";
```
Siemienik, line 4: That's actually breaking compatibility change, I don't see reason of doing that now. Also importing everything from exceljs works great for my projects. 
What kind of error have you got?

Siemienik (changes requested):
Thank you for your engagement, that's really great to have opportunity review your PR. 
I little misunderstood the reason of this PR, as all works good for my projects.

spec/typescript/exceljs.spec.ts
Siemienik: I would be glad for sharing me your test repo from screenshots.

alexandre-embie:
I actually got the same exact problem and this PR fixed it. To ensure proper development, I had to specifiy 
`"exceljs": "github:groozin/exceljs#fix/2548/fix-typescript-and-intellisense"`
in my package.json in the `devDependencies` 

Please fix this.

marisuxma:
Any update @Siemienik @groozin ?

spec/typescript/exceljs.spec.ts
groozin: I just pushed the repo here: https://github.com/groozin/exceljs-test 
That being said - using the `vscode` version 

```
Version: 1.85.1 (Universal)
Commit: 0ee08df0cf4527e40edc9aa28f4b5bd38bbff2b2
Date: 2023-12-13T09:48:06.308Z
Electron: 25.9.7
ElectronBuildId: 25551756
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
OS: Darwin x64 23.1.0
```

and `node 18` I cant reproduce anymore. Still using the version `4.30` of the `exceljs`. 
Maybe the recent `vscode` versions fixed the problem. 
But just in case I uploaded the repo I used to test.

groozin:
@marisuxma @alexandre-embie I tried both versions `4.3.0` and `4.4.0` and can't reproduce with the latest `vscode`. Meaning all is working well for me now. Lets see how to progress here.

alexandre-embie:
I'm using webstorm. I think that's where the problem is from, but your package is the only one I have that has this problem so there has to be something setup wrong

clecherbauer:
Hi, i have this issue too

Issue #514, opened 2018-03-13
Is there a way inserting values in columns.


miniil:
Hi,

I want to insert multiple values into one colum based on its header.

For example : 

var myValues = [1, 2, 3, 4];
var thisColumn = sheet.getColumn('idCol');
thisColumn.value = myValues?

and the result will be something like this
idCol
1
2
3
4

Thanks for your help.

vshreyas-tricog:
you can do it by array of object's
like 
var rows = [];
var row = {
   col_name : value
};
rows.push(row);
sheet_name.addRow(rows);

i hope this may work for you

miniil:
Hi,

Thanks for your response but it's not exactly what I have to do.

Because I open an existing file with many columns and I have to add a column at the end to append data.

Thanks for your help.

guyonroche:
@miniil I have just added a values property to Column (as of v1.1.0)
So now you can:
```
thisColumn.values = ['values', 'will', 'be', 'added', 'to', 'thisColumn'];
```

dmk1111:
@guyonroche looks like it is  readonly array as of 4.4.0: `values: ReadonlyArray<CellValue>;`
Any idea how to handle this properly now?

PR #2571, opened 2023-10-28
#514: Remove invalid readonly for Column.values

Siemienik:
## Summary

The `values` field for Column class is actually mutable.
https://github.com/exceljs/exceljs/issues/514#issuecomment-1782927522


pex:
This PR also accidentially removed the array type of `values`. Fixed in https://github.com/exceljs/exceljs/pull/2587

Issue #2267, opened 2023-05-10
[BUG] XYZ


techiesalman:
## 🐛 Bug Report

The `WorkbookWriter()` function in ExcelJS is overwriting RichText objects in multiple cells with the first RichText object.

Lib version: 4.3.0

## Steps To Reproduce

```import * as ExcelJS from 'exceljs';

const workbook = new ExcelJS.stream.xlsx.WorkbookWriter(options);
const worksheet = workbook.addWorksheet('Sheet 1');

const richText1 = [
  { font: { bold: true }, text: 'Bold text' },
  { text: ' and regular text.' }
];

const richText2 = [
  { font: { bold: true }, text: 'Just' },
  { text: ' another sentence' }
];

worksheet.getCell('A1').value = {richText: richText1}
worksheet.getCell('A2').value = {richText: richText2}

await workbook.commit();
```

## The expected behaviour:

Cell A1 contains the text: "Bold text and regular text." with "Bold text" in bold.
Cell A2 contains the text: "Just another sentence" with "Just" in bold

## Actual Behaviour:
Both cells A1 and A2 contain the same `RichText `object. They both display "Bold text and regular text." with "Bold text" in bold.
This issue prevents the use of different RichText objects in different cells when using `WorkbookWriter()`. Any insight or workaround would be appreciated.

![image](https://github.com/exceljs/exceljs/assets/37684962/d9872ddc-5545-4b1d-83f6-4e9ec4a3a452)

techiesalman:
```
const options = {
      filename: `${this.fileDirectory}\\${fileName}`,
      useStyles: true,
      useSharedStrings: false
    };
```

When `useSharedStrings` is false then it works.

Siemienik:
Thanks fot investing this It seems like comparing these cells by string values (toString) except rich text settings.

HelloYie:
+1， This seems to be a bug.

belanEu:
Hello!
Any updates on that bug?

belanEu:
When option `useSharedStrings: false` is set, text gets changed, but for some reason there are no richText styles applied in the result xlsx file.

belanEu:
@Siemienik Hi!
I see that PetrChalov has pushed some fixes. I'm wondering if those changes are going to be merged soon. 👉👈

belanEu:
[...]

PR #2588, opened 2023-11-13
Fix shared strings and richText

PetrChalov:
## Summary

Corrects the behavior of richText in SharedStrings

See https://github.com/exceljs/exceljs/issues/2267

## Test plan

Create a workbook with multiple richText values and enabled useSharedStrings.

### DIFF
--- a/lib/utils/shared-strings.js
+++ b/lib/utils/shared-strings.js
@@ -22,13 +22,25 @@ class SharedStrings {
   }
 
   add(value) {
-    let index = this._hash[value];
+    const key = this.getKeyFromValue(value);
+    let index = this._hash[key];
+
     if (index === undefined) {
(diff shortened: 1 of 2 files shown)


protobi-pieter:
This bug was independently rediscovered and fixed in the protobi fork by @gwkline ([protobi/exceljs#50](https://github.com/protobi/exceljs/pull/50), now landed as [protobi/exceljs#66](https://github.com/protobi/exceljs/pull/66) with regression tests). The two fixes take different approaches:

- **This PR (#2588):** modifies the hash key inline within `SharedStrings.add()`
- **gwkline's fix:** uses `SharedStringXform.toXml()` for the richText hash key (lazy instance per `SharedStrings`)

The regression tests in protobi#66 should apply to both approaches if useful for moving this forward. Original bug report: #2267.

PR #2649, opened 2024-01-04
fix: table column style bug and addTable colum type

jbuck-lineleap:
## Summary
A continuation of https://github.com/exceljs/exceljs/pull/1907 with an added test and updates typings to allow for styles to be set in the column definition of a table

## Test plan

In a project that is using exceljs as a dependency, i took the following steps:

1. Navigated to `node_modules/exceljs/lib/doc/table.js`
2. Made the change in the PR here `cell[key] = style[key]` => `cell.style[key] = style[key]`
3. Added a column with the following format to my table `{ name: 'Column Name', totalsRowFunction: 'sum', style: { numFmt: '0.00%' } }`
4. Regenerated the report and confirmed that the colum was formatted as `0.00%'`

https://github.com/exceljs/exceljs/blob/master/lib/doc/table.js#L183-L189

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1836,7 +1836,7 @@ export interface TableColumnProperties {
 	/**
 	 * Styles applied to the column
 	 */
-	style: Partial<Style>;
+	style?: Partial<Style>;
 }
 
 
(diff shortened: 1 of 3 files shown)


Siemienik (approved):
LGTM, merge on Merge Fest I / 2024

Issue #2678, opened 2024-02-05
[BUG] Table creation accepts empty row array, resulting in corrupted Excel file


georgbuehler:
## 🐛 Bug Report

When creating a table, it's possible to define the rows array, but still leave it empty. ExcelJS will not warn that this will be an issue, but it will result in the Excel file being corrupted.  One changed line of code in the validate() method in the Table class should resolve this issue.

Lib version: 9.4.2

## Steps To Reproduce

1. Run the Jest unit test (below) to a project with exceljs installed.
2. Open the generated EmptyRowsTableTest.xlsx file.
3. User will receive the following error message: "We found a problem with some content in 'BadTablenameTest.xlsx'. Do you want us to try to recover as much as we can? If you trust the source of this workbook, click Yes."
4. If you click "Yes" for the repairs, Excel will present the message: "Repaired Records: Table from /xl/tables/table1.xml part (Table)". However, the table will not be created in the file.

```javascript
import Excel from "exceljs";
import path from "path";

describe("addTable()", () => {
  it("should not allow an empty rows array", async () => {
    const excelDoc = new Excel.Workbook();
    const accountSheet = excelDoc.addWorksheet("Accounts");

    accountSheet.addTable({
      name: "Bob_s_Accounts",
      ref: "A1",
      headerRow: true,
      totalsRow: false,
      style: {
        theme: "TableStyleMedium2",
        showRowStripes: true,
      },
      columns: [
        { name: "Name", filterButton: true },
        { name: "Description", filterButton: true },
        { name: "Credit Limit", filterButton: true },
        { name: "Account Type", filterButton: true },
      ],
      rows: [], // EMPTY ROWS ARRAY -- currently this is permitted, but it will corrupt the Excel file
    });
    expect(excelDoc).toBeInstanceOf(Excel.Workbook);
    await excelDoc.xlsx.writeFile(
      path.join(__dirname, "../../data/EmptyRowsTableTest.xlsx")
    );
  });
});

```

## The expected behaviour:

When the developer tries to create a Table without populating the rows array, a runtime error should result: "Table must have row definitions."

## Possible solution (optional, but very helpful):

Update the validate() function to check that the rows array is not only defined by also populated:
```javascript
assert(table.rows && table.rows.length > 0, 'Table must have row definitions');
```

PR #2680, opened 2024-02-06
Fix for Bug 2678: Table creation allows empty array of rows

georgbuehler:
## Summary
This is a fix for Bug #2678. It adds an additional test condition to the validate() method on the Table class, to ensure that the rows array is actually populated.

## Test plan

The PR includes a unit test file with the appropriate tests to confirm a complete set of table options works fine, while not populating the rows array results in a run-time error. I've run the full set of Mocha unit tests in my local system and all tests passed.

One line of code changed in [Table.js](https://github.com/exceljs/exceljs/blob/master/lib/doc/table.js)

### DIFF
--- a/lib/doc/table.js
+++ b/lib/doc/table.js
@@ -151,7 +151,7 @@ class Table {
     };
     assert(table.ref, 'Table must have ref');
     assert(table.columns, 'Table must have column definitions');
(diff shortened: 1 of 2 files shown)


abeel50 (approved):
Great contribution

PR #2687, opened 2024-02-13
[Breaking] Replace unzipper with yauzl-promise

### LINKED ISSUE #2686: no licensing on one of exceljs dependencies
Not sure if this is the place to ask, but I'm using exceljs, and this uses [buffers](https://www.npmjs.com/package/buffers) which is used by binary of which node-unzipper references. Security alarms went off stating that there is no license and the github accounts have been deleted. I'm wondering what your thoughts are on this.

pnappa:
## Summary

Unzipper had a transitive dependency on a package that does not have a license. This means that every user is potentially violating the law in using it. Personally, I think it's better to stick to following it, so I've replaced the root dependency with a modern (and updated) alternative.

However, yauzl-promise requires node v16 and above! I think there's some bikeshedding to be done for whether this is worthwhile or not.

This is a breaking change, as the minimum node engine has been bumped from v10 to v16. Note that the engines before this claimed v8 was supported, but the transitive dependencies required >=10. I believe this is an okay change to merge, as anything older than v18 is not supported anymore. https://endoflife.date/nodejs

## Test plan

The test suite passes locally (on my machine).

Resolves https://github.com/exceljs/exceljs/issues/2686

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [10.x, 12.x, 14.x, 16.x, 17.x, 18.x, 19.x, 20.x]
+        node-version: [16.x, 17.x, 18.x, 19.x, 20.x]
         os: [ubuntu-latest, macOS-latest, windows-latest]
     runs-on: ${{ matrix.os }}
 
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-package-lock=false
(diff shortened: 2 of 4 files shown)


jjshinobi:
This change will also resolve https://security.snyk.io/vuln/SNYK-JS-INFLIGHT-6095116 due to the transitive dependency
exceljs@4.4.0 › unzipper@0.10.14 › fstream@1.0.12 › rimraf@2.7.1 › glob@7.2.3 › inflight@1.0.6. Can we please bump this PR for review?

pnappa:
Another alternative is to implement what's described here in https://github.com/exceljs/exceljs/issues/2703 . I probably reckon this is a better solution. It's supported in NodeJS (>=18), which is the minimum version that this PR requires anyway.

Siemienik:
Hi everyone,  
I’ve released  a version with PR [Request to bump version of unzipper to eliminate license concerns](https://github.com/exceljs/exceljs/pull/2744) as a pre-release version: `4.4.1-prerelease.0`.  

If no issues are reported with this version, I plan to officially release it as `4.4.1` soon.

Issue #2689, opened 2024-02-13
[BUG] Worksheets with a lot of merged cells take forever to parse


aorsten:
## 🐛 Bug Report
Files with too many merged cells will not load - they take too long.

Lib version: 4.4.0

## Steps To Reproduce

1. Create an Excel book with a worksheet.
2. In the worksheet, add a large amount of rows where one cell is merged on each row. See attached file. 
[toomanymerges.xlsx](https://github.com/exceljs/exceljs/files/14271763/toomanymerges.xlsx)
3. Try to parse the workbook using:
```
  const buffer = fs.readFileSync(filename);
  const workbook = new Workbook();
  console.time('Buffer');
  await workbook.xlsx.load(buffer);
```
4. Experience how it takes a long long time to parse.

## The expected behaviour:

It should not be so slow. The attached file takes 1.5 minutes to parse in my Node environment. I have encountered worse examples from users of our product, and it runs in the browser, so those instances have just never finished.

## Possible solution (optional, but very helpful):

I have noticed that the slow culprit is the `_mergeCellsInternal` function inside `worksheet.js`, specifically the check that cells aren't already merges. Which makes sense, because it is built with n squared loops. See my discussion here: https://github.com/exceljs/exceljs/discussions/2568 

```javascript
// worksheet.js, ._mergeCellsInternal
    // check cells aren't already merged
    _.each(this._merges, merge => {
      if (merge.intersects(dimensions)) {
        throw new Error('Cannot merge already merged cells');
      }
    });
```

The above code is the culprit. When I comment out this code, parsing is very fast again. Is there a way this could be more efficient? Is it even necessary? Is it necessary on first parse of the document?

aorsten:
Question: Rather than loop every other merge to see if there's a collision, couldn't the following:

```
    // apply merge
    const master = this.getCell(dimensions.top, dimensions.left);
    for (let i = dimensions.top; i <= dimensions.bottom; i++) {
      for (let j = dimensions.left; j <= dimensions.right; j++) {
        // merge all but the master cell
        if (i > dimensions.top || j > dimensions.left) {
          this.getCell(i, j).merge(master, ignoreStyle);
        }
      }
    }
```
Check the cells in question and see if they are already part of a merge? As such there are only n-merged cells included in the check?

aorsten:
My attempt at a fix in a PR: https://github.com/exceljs/exceljs/pull/2691

hubyhub:
[...]

PR #2691, opened 2024-02-15
fix: inefficient merge check for large amount of merged cells within …

aorsten:
When there are a lot of merged cells within a worksheet, e.g. 30.000 merged cells (e.g. the same 2 cells are merged in every row in a 30.000 row table), parsing mergeCells is increasingly slow. This is due to the inefficient conflict check within `_mergeCellsInternal(dimensions, ignoreStyle)` in worksheet.js.

## Summary

My motivation is to let these big files function, rather than load forever. I believe the massive check is unnecessary. Rather than check every other merged range, can't you just see if any of the cells in the new range are merged, and then throw an error?

I do not know the original motivation for the `throw new Error('Cannot merge already merged cells');` error, and whether it always breaks the program, or if the error is caught somewhere. Hopefully the new implementation is much more efficient without breaking anything.

## Test plan

See the issue https://github.com/exceljs/exceljs/issues/2689
I have attached a file there. Loading that file goes from 1.5 minutes in node JS to only seconds. In the browser that file just will not load.

### DIFF
--- a/lib/doc/worksheet.js
+++ b/lib/doc/worksheet.js
@@ -613,20 +613,23 @@ class Worksheet {
   }
 
   _mergeCellsInternal(dimensions, ignoreStyle) {
-    // check cells aren't already merged
-    _.each(this._merges, merge => {
-      if (merge.intersects(dimensions)) {
-        throw new Error('Cannot merge already merged cells');
-      }
-    });
-
     // apply merge
     const master = this.getCell(dimensions.top, dimensions.left);
+    // check cells aren't already merged
+    if (master.isMerged()) {
+      throw new Error('Cannot merge already merged cells');
+    }
+    let cell;
     for (let i = dimensions.top; i <= dimensions.bottom; i++) {
       for (let j = dimensions.left; j <= dimensions.right; j++) {
         // merge all but the master cell
         if (i > dimensions.top || j > dimensions.left) {
-          this.getCell(i, j).merge(master, ignoreStyle);
+          cell = this.getCell(i, j);
+          // check cells aren't already merged
+          if (cell.isMerged()) {
+            throw new Error('Cannot merge already merged cells');
+          }
+          cell.merge(master, ignoreStyle);
         }
       }
     }


aorsten:
Fixes https://github.com/exceljs/exceljs/issues/2689

lib/doc/worksheet.js
```diff
@@ -613,20 +613,23 @@ class Worksheet {
...
-    _.each(this._merges, merge => {
-      if (merge.intersects(dimensions)) {
-        throw new Error('Cannot merge already merged cells');
-      }
-    });
-
     // apply merge
     const master = this.getCell(dimensions.top, dimensions.left);
+    // check cells aren't already merged
+    if (master.isMerged()) {
```
markygab, line 619: isMerged is not a function, its  a getter so you need this.

```suggestion
    if (master.isMerged) {
```

lib/doc/worksheet.js
```diff
@@ -613,20 +613,23 @@ class Worksheet {
...
+    }
+    let cell;
     for (let i = dimensions.top; i <= dimensions.bottom; i++) {
       for (let j = dimensions.left; j <= dimensions.right; j++) {
         // merge all but the master cell
         if (i > dimensions.top || j > dimensions.left) {
-          this.getCell(i, j).merge(master, ignoreStyle);
+          cell = this.getCell(i, j);
+          // check cells aren't already merged
+          if (cell.isMerged()) {
```
markygab, line 629: Same here

markygab (changes requested):
Hi there, you change will indeed make it much faster, but you've got a couple of errors that prevent it from running atm

lib/doc/worksheet.js
aorsten: @markygab I'm sorry, I didn't notice your review until now, and I am super busy at work and home with something else. Are you or anyone else able to apply your suggestions and test them? I recall that I did test my functionality in my own application, and it seemed to work, but perhaps I just didn't notice there was a bug somewhere.

aorsten: Or, if you need me to, I could use the github UI to apply suggestions, so you could review again. Let me know. :)

3ximus:
I've applied the suggested comments on this PR: https://github.com/exceljs/exceljs/pull/2920 (preserving the author's commit)
So you could discard this one so you could apply the other one instead.
Cheers.

PR #2697, opened 2024-02-19
Add type DataValidationType

gregfenton:
## Summary

Being able to use `as DataValidationType` is cleaner than having to use `as 'list' | 'whole' | 'decimal' | 'date' | 'textLength' | 'custom'` everywhere.

This new type also makes the pattern consistent with the `DataValidationOperator` type also used in `DataValidation`.

## Test plan

No functional code changes.

Added a new type (DataValidationType) and used it in one other type definition (DataValidation).


mohqarmout:
Is there any update on this? , I think it's going to be a good type to have :)

PR #2698, opened 2024-02-20
Issue: style.xml has [Object object] as formatCode

jmatth11:
## Summary

Somehow the excel file is in a state where exceljs does not handle extracting the numFmt of a DXF conditional format properly and instead has it set to an object.

This results in the generate style.xml file to have a formatCode of `[Object object]` inside which corrupts the file.

This fix is to check if the numFmt is of the appropriate string type and if not, corrects it by extracting the formatCode before moving forward on the prepare step of writing.

## Test plan

This is the test Excel file I used which has a conditional format on A1:C1 which conditional changes the fill color to yellow and the format to be a custom number format if A1 equals "NO".

[CF1.xlsx](https://github.com/exceljs/exceljs/files/14351092/CF1.xlsx)

I setup a simple `main.js` to read in the test Excel file and write a new file `temp.xlsx`.

```
const ExcelJS = require("exceljs");
const workbook = new ExcelJS.Workbook();
workbook.xlsx.readFile("CF1.xlsx").then(() => {
  workbook.xlsx.writeFile("temp.xlsx");
});
```

With current master branch the result of opening the generated file is this:

original_error
original_error_repair_report

After repair conditional formatting is removed.

original_error_cf_removed

Screenshot of xl/style.xml file with the error:

original_xml_file

After the fix the file is generated properly and conditional formatting works as expected.

after_fix

Screenshot of xl/style.xml with fixed format code:

after_fix_xml_file

I'm not really sure what is supposed to go here.

### DIFF
--- a/lib/xlsx/xform/style/styles-xform.js
+++ b/lib/xlsx/xform/style/styles-xform.js
@@ -345,6 +345,11 @@ class StylesXform extends BaseXform {
 
   addDxfStyle(style) {
     if (style.numFmt) {
+      // Check if numFmt is not a string and if instead it is an object with the formatCode
+      // extract the formatCode out as the numFmt
+      if (!(style.numFmt instanceof String) && style.numFmt.hasOwnProperty("formatCode")) {
+          style.numFmt = style.numFmt.formatCode;
+      }
       // register numFmtId to use it during dxf-xform rendering
       style.numFmtId = this._addNumFmtStr(style.numFmt);
     }
--- a/lib/xlsx/xform/style/styles-xform.js
+++ b/lib/xlsx/xform/style/styles-xform.js
@@ -345,8 +345,8 @@ class StylesXform extends BaseXform {
 
   addDxfStyle(style) {
     if (style.numFmt) {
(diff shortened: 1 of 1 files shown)


lib/xlsx/xform/style/styles-xform.js
```diff
@@ -345,6 +345,11 @@ class StylesXform extends BaseXform {
 
   addDxfStyle(style) {
     if (style.numFmt) {
+      // Check if numFmt is not a string and if instead it is an object with the formatCode
+      // extract the formatCode out as the numFmt
```
patra-nash-stokes:
```suggestion
      // Check if numFmt is not a string and if it is an object with the formatCode property.
      // If both are true, set numFmt's value equal to the value of the formatCode property.
```

patra-nash-stokes (approved):

thestuckster:
lgtm

Ks89:
I found the same problem, please could you merge and publish a patch release.
Thanks.

alex-mazzariol:
We're finding this issue too when updating an existing excel file that has conditional formatting for some cells. :+1:

PR #2710, opened 2024-02-28
fix: add proper version control to deps

haikal-handamara-x150s:
## Summary

Package `tmp` release new version [0.2.2](https://github.com/raszi/node-tmp/compare/v0.2.1...v0.2.2) which changes the minimum node version from `>=8.17.0` to `>=14` which caused any packages that dependent to `tmp` may broken. This commit lock the `tmp` version to safe version which would not break any application that depends to `exceljs@4.4.0` but running in older NodeJS version (<14).

## Test plan

Normal CI/CD test, but on NodeJS <14

Nothing


pnappa:
I really hope you're not using such an old version of NodeJS in a production environment. Version 8 hasn't had any fixes for four years. https://endoflife.date/nodejs 

Please update to a newer version (e.g. 18 or 20). You are at risk of security vulnerabilities.

Issue #2730, opened 2024-03-27
[BUG] When using the addImage method and specifying the cell width with tl and br after changing the cell width, the intended width is not set.


RyokoKassai:
## 🐛 Bug Report

When using the addImage method and specifying the cell width with tl and br after changing the cell width, the intended width is not set. The image is displayed with a smaller cell width than specified.

Is there a way to specify the "br" while changing the cell width? 
Please note that I also want to set the editAs property at the same time, so specifying it with br instead of ext is necessary.

Lib version: 4.4.0

## Steps To Reproduce

■javascript source

    const workbook = new ExcelJS.Workbook();

    const worksheet = workbook.addWorksheet('My Sheet');
    worksheet.columns = [
        { header: 'Id', key: 'id', width: 10 },
        { header: 'Name', key: 'name', width: 32 },
        { header: 'test', key: 't1', width: 32 },
        { header: 'test2', key: 't2', width: 32 },
        { header: 'test3', key: 't3', width: 32 },
      ];

    // read image files
    const imageId = workbook.addImage({
        filename: 'img_7483.jpg',
        extension: 'jpg',
    });
    
    worksheet.addImage(imageId,{
        tl:{col:2,row:2},
        br:{col:3.8,row:3.5},
        editAs:'twoCell'
    });
    
    // save Excel file
    workbook.xlsx.writeFile(outputfile)
        .then(() => {
            console.log('completed!');
        })
        .catch(error => {
            console.error('an error ocurred.', error);
        });

## The expected behaviour:

Expected: The specified "br" is applied to the changed cell width, and the image is displayed.
Actual: The image is displayed with a smaller cell width than specified.
[out_otamesi_gazo.xlsx](https://github.com/exceljs/exceljs/files/14766158/out_otamesi_gazo.xlsx)

wwwxy80s:
I read the source code of exceljs and consulted some information. I found a way to accurately add images through the br attribute, but I don’t know if it is universal. I would like to share some information I found.

When calling addImage using the br attribute, you need to determine the two anchor points of the upper left corner and the lower right corner of the image.

[...]

Issue #1747, opened 2021-06-16
[BUG] Type '{ col: number; row: number; }' is missing the following properties from type 'Anchor': nativeCol, nativeColOff, nativeRow, nativeRowOff, and 3 more.


jakehockey10:
## 🐛 Bug Report

Lib version: 4.2.1

## Steps To Reproduce

```javascript
worksheet.addImage(imageId, {
  tl: { col: 0, row: 31 },
  br: { col: 15, row: 31 },
});
```

## The expected behaviour:

In TypeScript projects (like my Angular project) will not run if I use the example syntax shown in the README: 
![image](https://user-images.githubusercontent.com/1821383/122255507-6d08c780-ce8b-11eb-8cbc-2458184aacdd.png)

I get the following errors and cannot build because the interface requires all properties to be present.

![image](https://user-images.githubusercontent.com/1821383/122255573-7e51d400-ce8b-11eb-90a7-522ae373f4c2.png)

## Possible solution (optional, but very helpful):

```javascript
// You could make those properties optional by putting a `?` before the semicolon defining the properties type to be optional.  

export interface Anchor {
  nativeCol?: number;
  nativeColOff?: number;
  nativeRow?: number;
  nativeRowOff?: number;
  ....
}

// Or you could use a Partial of Anchor in the method declaration for `addImage`:

Partial<Anchor>
```

jakehockey10:
You can get around this by doing the following, but I couldn't find an example of this in the documentation:

```typescript
worksheet.addImage(imageId, {
    tl: new Anchor({ col: 0, row: 31 }),
    br: new Anchor({ col: 15, row: 31})
});
```

jakehockey10:
I should clarify.  This is not a solution.  This let's my project transpile.  However, I get an error: `exceljs__WEBPACK_IMPORTED_MODULE_0__.Anchor is not a constructor`.

This is what my import statement looks like:

```typescript
import { Anchor, Row, Workbook, Worksheet } from 'exceljs';
```

mjiskinng:
Hi @jakehockey10 what is the solution to this problem. Please i am going crazy with this image stretch issue.. Nothing is working. Thanks in advance if your solution can make my life easier.

jakehockey10:
So far? To not write this code. 

I was really hoping I'd get a response. But I have no solution to this

Rob4226:
Any updates??

Emiya0306:
Is there any updates??? I got the same error.
[...]

PR #2783, opened 2024-06-26
:memo: Fix errors in document about image embedding

Cat7373:
## Summary

In the current documentation, the Chinese and English documents are inconsistent. One uses oneCell, the other uses oneCells. One says absolute is the default value, the other says oneCell is the default value. In addition, passing undefined in the English document is invalid because it will match the default value oneCell.

Fixed the inconsistency between the image embedding part in the document and the actual code. Please refer to the following Issue:

https://github.com/exceljs/exceljs/issues/2730

https://github.com/exceljs/exceljs/issues/1747

Please also refer to the documentation from Microsoft:

https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.drawing.spreadsheet.editasvalues?view=openxml-3.0.1

## Test plan
From my project code, it works correctly:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -2042,24 +2042,24 @@ Fractions of cells can be specified by using floating point numbers, e.g. the mi
 ```javascript
 // insert an image over part of B2:D6
 worksheet.addImage(imageId2, {
-  tl: { col: 1.5, row: 1.5 },
-  br: { col: 3.5, row: 5.5 }
+  tl: new Anchor(worksheet, { col: 1.5, row: 1.5 }),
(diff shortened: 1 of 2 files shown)


protobi-pieter:
Thanks for this important documentation fix! 🎉

We've adopted this change into our actively maintained fork at https://github.com/protobi/exceljs. Your contribution fixes real user confusion about image embedding behavior and aligns the docs with Microsoft's official OpenXML documentation.

Your work is appreciated and credited in commit 30b9971.

Thanks for taking the time to improve the documentation!

Issue #2790, opened 2024-07-07
[BUG] stream xlsx dropping worksheets


LarryKen:
## 🐛 Bug Report

The stream namespace for xlsx workbook reader does not correctly parse large numbers of worksheets > 100. This is due to the way it is unzipping the file, it is reading the entries from the unzipper parse however it is not correctly draining each entry resulting in the stream halting [node-unzipper](https://github.com/ZJONSSON/node-unzipper?tab=readme-ov-file#parse-zip-file-contents:~:text=If%20you%20do%20not%20intend%20to%20consume%20an%20entry%20stream%27s%20raw%20data%2C%20call%20autodrain()%20to%20dispose%20of%20the%20entry%27s%20contents.%20Otherwise%20the%20stream%20will%20halt.). The iterateStream util does not call autodrain for each data when reading the unzipper stream.

Lib version: 4.4.0

## Steps To Reproduce

```javascript
    const fileStream = createReadStream(PATH_TO_WORKBOOK_WITH_100_WORSHEETS);
    const workbookReader = new ExcelJS.stream.xlsx.WorkbookReader(fileStream, {
            worksheets: 'emit'
    });
    let count = 0;
    for await (const worksheetReader of workbookReader) {
        console.log('worksheet', worksheetReader.name);
        count+=1;
    };
```

## The expected behaviour:

Should consistently read all worksheets of a xlsx file.

elouie99:
Are there any guarantees that the workbookReader would return worksheets in the order they appear in the workbook? I ran a test and it seems like the order is not guaranteed. The total number of worksheets returned appear correct.

PR #2791, opened 2024-07-07
Issue 2790/xlsx stream missing worksheets

LarryKen:
## Summary

Provides a integration test case for [issue 2790](https://github.com/exceljs/exceljs/issues/2790) as well as a possible solution.

The stream namespace for xlsx workbook reader does not correctly parse large numbers of worksheets > 100. This is due to the way it is unzipping the file, it is reading the entries from the unzipper parse however it is not correctly draining each entry resulting in the stream halting [node-unzipper](https://github.com/ZJONSSON/node-unzipper?tab=readme-ov-file#parse-zip-file-contents:~:text=If%20you%20do%20not%20intend%20to%20consume%20an%20entry%20stream%27s%20raw%20data%2C%20call%20autodrain()%20to%20dispose%20of%20the%20entry%27s%20contents.%20Otherwise%20the%20stream%20will%20halt.). The iterateStream util does not call autodrain for each data when reading the unzipper stream.

## Test plan

This is the test xlsx file that has 188 worksheets.
[WORKBOOK_WITH_188_SHEETS.xlsx](https://github.com/LarryKen/exceljs-fork/blob/issue-2790/xlsx-stream-missing-worksheets/spec/integration/data/WORKBOOK_WITH_188_SHEETS.xlsx)

I've created a [test case](https://github.com/LarryKen/exceljs-fork/blob/issue-2790/xlsx-stream-missing-worksheets/spec/integration/issues/issue-2790-xlsx-stream-read-large-worksheet-count.spec.js) that will read the file and check whether the count of worksheets emitted is equal to the count of worksheets in the file.

Running `npm run test:integration` get 194 tests passing 2 failing (The 2 failing tests are also failing on master).

image


elouie99:
Are there any guarantees that the workbookReader would return worksheets in the order they appear in the workbook? I ran a test and it seems like the order is not guaranteed. The total number of worksheets returned appear correct.

In addition, each worksheet returned doesn't appear to return the correct row data if I iterate through the rows and add rows to a new worksheet. If you add some sample rows to a few worksheets in the workbook and run this code, it should reproduce.



    async function printWorkbook() {
        const outputWorkbook = new ExcelJS.Workbook();
        let combinedWorksheet = outputWorkbook.addWorksheet('Test worksheet');    
        const filePath = 'WORKBOOK_WITH_188_SHEETS.xlsx';
        const workbookReader = new ExcelJS.stream.xlsx.WorkbookReader(filePath, {
          worksheets: 'emit',
        });
  
        for await (const worksheet of workbookReader) {
        console.log('===> worksheet:', worksheet.name);
            for await (const row of worksheet) {
                combinedWorksheet.addRow(row.values).commit();
            }
        }
        await outputWorkbook.xlsx.writeFile('combined.xlsx');
    }

    printWorkbook().catch(console.error);

LarryKen:
There are no guarantee that the worksheets are returned in the way they appear in the workbook. I believe the order is effectively set by how unzipper extracts the xlsx. I believe the order on this branch is the same as on `master`.

Also regarding the rows not return the correct data. I think that is not an issue introduced by my changes. I am able to replicate the same results on `master` as on my branch.

PR #2803, opened 2024-08-01
Fix corrupted file with conditional formatting and hyperlinks

TheAsda:
## Summary

I faced the issue with stream writer creating corrupted file. Through some experiments and inspecting differences between the file created by the library and handmade file using Microsoft Excel I found out that the corruption appears when you use conditional formatting together with hyperlinks. The issue is that the order of blocks in the worksheet is wrong. This pull request solves this issue.

## Test plan

First we create file using Microsoft Excel with links and some conditional formatting eg. `=MOD(ROW();2)=0` to color every odd row:
image
Then we inspect the file and see that the order in sheet is: first conditional formatting, second hyperlinks
image
After that we create similar file using the library:
```js
const ExcelJS = require("exceljs");
const fs = require("fs");

const fileStream = fs.createWriteStream("output.xlsx");

const workbook = new ExcelJS.stream.xlsx.WorkbookWriter({
  stream: fileStream,
  useSharedStrings: false,
  useStyles: true,
});
const worksheet = workbook.addWorksheet("Sheet1");
for (let index = 0; index < 4; index++) {
  worksheet.addRow([
    "test",
    {
      text: "Link",
      hyperlink: "https://google.com",
      tooltip: "https://google.com",
    },
  ]);
}

worksheet.addConditionalFormatting({
  ref: `A1:B4`,
  rules: [
    {
      priority: 1,
      type: "expression",
      formulae: ["MOD(ROW(),2)=0"],
      style: {
        fill: {
          type: "pattern",
          pattern: "solid",
          bgColor: "FFEDEDED",
        },
      },
    },
  ],
});

worksheet.commit();
workbook.commit();

```

If we open the resulting file with Excel we see the error:

image

If we inspect the underlying xml file we see that the order is different:

image

So then we can change the order in `worksheet-writer.js` and try the script again:

image

The file generates successfully and opens in Excel:

image

### DIFF
--- a/lib/stream/xlsx/worksheet-writer.js
+++ b/lib/stream/xlsx/worksheet-writer.js
@@ -237,8 +237,8 @@ class WorksheetWriter {
     // for some reason, Excel can't handle dimensions at the bottom of the file
     // this._writeDimensions();
 
-    this._writeHyperlinks();
     this._writeConditionalFormatting();
+    this._writeHyperlinks();
     this._writeDataValidations();
     this._writeSheetProtection();
     this._writePageMargins();


dezmunt1 (approved):

J43fura:
I see this PR and many others are not merged,
Did you fork this repo or migrate to another package? If so what did you choose to use?

TheAsda:
We used yarn patch functionality

J43fura:
Thank you

protobi-pieter:
👋 We've adopted this fix in our [@protobi/exceljs](https://www.npmjs.com/package/@protobi/exceljs) fork!

✅ **Status**: Applied and tested successfully
📦 **Version**: Included in 4.4.0-protobi.2
🔗 **Our issue**: https://github.com/protobi/exceljs/issues/25
💾 **Our commit**: https://github.com/protobi/exceljs/commit/0509050

This fix resolves file corruption when using conditional formatting and hyperlinks together. We're tracking this and 12 other upstream PRs, and will sunset our fork once they're merged upstream.

Thank you for this important fix! 🙏

PR #2840, opened 2024-10-21
fix: styles from cellStyleXfs are not preserved (#2930)

christianvoigt:


christianvoigt:
sorry, I accidentally created this. I still need to fix tests before this is ready for review. Closing this for now.

PR #2849, opened 2024-11-21
feat: support web-native streams for read/write methods

### LINKED ISSUE #1228: Support Stream writing in browser
## 🚀 Stream writing should be possible in browser

If you export large excel file in browser, it crashes on memory. Stream writing which is supported in NodeJS should be supported in browser.

### LINKED ISSUE #2753: [Q] What's the status to support Web Stream ?
Web Stream are supported in modern browsers.

Do we have plan to support it ?  Are there technical obstacles to implement?

see: #1228, after four years it's still open, no response from team members.

lionel-rowe:
Fixes https://github.com/exceljs/exceljs/issues/1228
Fixes https://github.com/exceljs/exceljs/issues/2753

## Summary

Add support for `read(<web-native ReadableStream instance>)` and `write(<web-native WritableStream instance>)`. Both have excellent support in modern browsers and are also implemented in NodeJS, Deno, and Bun.

## Test plan

See `spec/unit/xlsx/write-writable-stream.spec.js` and changes to `spec/typescript/exceljs.spec.ts`.

N/A (updated typings are reflected in diffs to runtime code + tests).

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   },
   "main": "./excel.js",
   "browser": "./dist/exceljs.min.js",
-  "types": "./index.d.ts",
+  "types": "./exceljs.d.ts",
   "files": [
     "dist",
     "lib",
@@ -26,12 +26,12 @@
     "README.md",
     "README_zh.md",
     "index.ts",
-    "index.d.ts"
+    "exceljs.d.ts"
   ],
   "scripts": {
     "test": "npm run test:full",
     "test:es5": "export EXCEL_BUILD=es5 && npm run test:full",
-    "test:full": "npm run build && npm run test:unit && npm run test:integration && npm run test:end-to-end && npm run test:jasmine",
+    "test:full": "npm run build && npm run test:unit && npm run test:integration && npm run test:end-to-end && npm run test:jasmine && npm run test:typescript",
     "test:version": "npm run build && npm run test:unit && npm run test:integration && npm run test:end-to-end && npm run test:browser && npm run test:dist",
     "test:all": "npm run test:native && npm run test:es5",
     "test:native": "npm run test:full",
--- a/spec/typescript/exceljs.spec.ts
+++ b/spec/typescript/exceljs.spec.ts
@@ -1,7 +1,10 @@
 import 'regenerator-runtime/runtime';
 
 import { expect } from 'chai';
-import ExcelJS from '../../index';
+import _ExcelJS from '../../index';
+import type * as ExcelJSModule from '../../exceljs';
+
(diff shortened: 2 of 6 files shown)


spec/typescript/exceljs.spec.ts
```diff
@@ -1,7 +1,10 @@
 import 'regenerator-runtime/runtime';
 
 import { expect } from 'chai';
-import ExcelJS from '../../index';
+import _ExcelJS from '../../index';
+import type * as ExcelJSModule from '../../exceljs';
+
+const ExcelJS = _ExcelJS as unknown as typeof ExcelJSModule;
```
lionel-rowe, line 7: Type checking wasn't currently being picked up by ts-node due to `../../index` resolving to the untyped index files rather than `../../index.d.ts`.

https://stackoverflow.com/questions/62490423/consume-index-d-ts-declaration-file-in-source-code

> It looks like you have a collision in your module resolution.
>
> If you have a file called index.ts, then index.d.ts won't be automatically imported.
> 
> TypeScript will automatically import .d.ts files, UNLESS their name collides with other files, where the rules for what files will collide are somewhat described here: https://www.typescriptlang.org/docs/handbook/module-resolution.html .

For that reason, I've renamed `index.d.ts` to `exceljs.d.ts` (and updated the corresponding package.json fields etc).

It still needs to be imported in this weird way for the test though. Possibly there's a better fix, but I couldn't figure one out.

spec/typescript/exceljs.spec.ts
```diff
@@ -24,17 +27,21 @@ describe('typescript', () => {
     ws.getCell('A1').value = 7;
 
     const wb2 = new ExcelJS.Workbook();
-    const stream = wb2.xlsx.createInputStream();
```
lionel-rowe, line 27: This test was previously failing due to removal of `createInputStream`, so I've overwritten the existing test case rather than adding a new one.

package.json
```diff
@@ -26,12 +26,12 @@
...
     "README_zh.md",
     "index.ts",
-    "index.d.ts"
+    "exceljs.d.ts"
   ],
   "scripts": {
     "test": "npm run test:full",
     "test:es5": "export EXCEL_BUILD=es5 && npm run test:full",
-    "test:full": "npm run build && npm run test:unit && npm run test:integration && npm run test:end-to-end && npm run test:jasmine",
+    "test:full": "npm run build && npm run test:unit && npm run test:integration && npm run test:end-to-end && npm run test:jasmine && npm run test:typescript",
```
lionel-rowe, line 34: Added TS tests to `test:full`, which should now work as they no longer contain a failing test case.

PR #2983, opened 2025-10-08
Add ImageEditAs type and update addImage method to support 'twoCell' option

hshoja:
## Summary

This PR adds support for the `twoCell` `editAs` option when adding images to Excel worksheets. This feature allows images to be moved and sized with cells, providing better integration with Excel's filtering and row manipulation features.

This addresses the issue discussed in [GitHub Discussion #2648](https://github.com/exceljs/exceljs/discussions/2648#discussioncomment-8133150) where the images remain visible when rows containing them are filtered or removed. The current library only supports `oneCell` (default) and `absolute` positioning, but Excel's native `twoCell` option provides the expected behavior for images that should move and resize with their containing cells.

- [x] All existing tests pass
- [x] Added new integration test for `twoCell` functionality
- [x] Manual testing completed with example code

## Test plan

See working example demonstrating the `twoCell` functionality: [ExcelJS twoCell Example](https://gist.github.com/hshoja/dd5f9ab63ce60a9b94646150b831d37f)

```javascript
worksheet.addImage(imageId, {
  tl: { col: 1, row: 1 },
  br: { col: 2, row: 2 },
  editAs: "twoCell",
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -2054,6 +2054,7 @@ It can have one of the following values:
 | --------- | ----------- |
 | undefined | It specifies the image will be moved and sized with cells |
 | oneCell   | This is the default. Image will be moved with cells but not sized |
+| twoCell   | Image will be moved and sized with cells |
 | absolute  | Image will not be moved or sized with cells |
 
 ```javascript
--- a/index.d.ts
+++ b/index.d.ts
@@ -400,6 +400,8 @@ export type CellValue =
 
 	export type CommentEditAs = 'twoCells' | 'oneCells' | 'absolute';
 
+	export type ImageEditAs = 'twoCell' | 'oneCell' | 'absolute';
+
 	export interface Comment {
 		texts?: RichText[];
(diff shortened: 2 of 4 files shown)


IamMille:
@hshoja is this suppose to keep the aspect ratio also?

hshoja:
> @hshoja is this suppose to keep the aspect ratio also?

By default, the "lock aspect ratio" settings are applied in all cases within Excel.

Screenshot 2025-12-03 at 10 52 46



 I introduced `twoCells` to support the first option, which is "move and resize with cells." Since, I found that the "undefined" value isn't functioning properly in the current implementation. 

The **editAs** property in the library is defined as follows: 

```
undefined: Image will be moved and sized with cells
oneCell: Image will be moved with cells but not sized (default)
absolute: Image will not be moved or sized with cells
```
 
You might want to try using `oneCell`, but I'm not entirely sure if it will work as expected.

PR #3056, opened 2026-06-23
fix: TypeError when shared string has empty <t> followed by <r> rich …

yuanliwei:
## Summary

When a `<si>` element in `sharedStrings.xml` contains both an empty `<t></t>` and a `<r>` rich text element, `SharedStringXform.parseClose()` assigns the empty string `''` to `this.model` from the `<t>` parser. Then when the subsequent `<r>` element closes, it tries to set `this.model.richText = []`, which throws:

```
TypeError: Cannot create property 'richText' on string ''
```

This occurs in real-world xlsx files generated by tools that produce `<si><t></t><r>...</r></si>` structures (e.g. AI-generated Excel exports).

The fix adds a fallback `|| {}` when assigning `this.parser.model` in the `case 't'` branch, ensuring `this.model` is always an object before the `<r>` branch tries to set properties on it.

## Test plan

A minimal reproduction xlsx file is included (`spec/integration/data/bug-repro-empty-shared-string.xlsx`) with a sharedStrings.xml containing:
```xml
<si><t></t><r><rPr><b/></rPr><t>bold text</t></r></si>
```

Before fix:
```
  1 failing:
  TypeError: Cannot create property 'richText' on string ''
```

After fix:
```
  ✓ should load without TypeError
```

## Related to source code

[`lib/xlsx/xform/strings/shared-string-xform.js#L83-L84`](https://github.com/exceljs/exceljs/blob/master/lib/xlsx/xform/strings/shared-string-xform.js#L83-L84) — the `case 't'` branch assigns `this.parser.model` directly, which can be an empty string when `<t></t>` is parsed. The subsequent `case 'r'` at [line 75-81](https://github.com/exceljs/exceljs/blob/master/lib/xlsx/xform/strings/shared-string-xform.js#L75-L81) expects `this.model` to be an object.


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "exceljs",
+  "name": "@yuanliwei/exceljs",
```
gameroman, line 2: .

spec/unit/xlsx/xform/strings/shared-string-xform.spec.js
```diff
@@ -33,8 +33,7 @@ const expectations = [
         },
       ],
     },
```
gameroman, line 35: .

spec/integration/issues/issue-1328-xlsx-worksheet-reader-date.spec.js
```diff
@@ -24,7 +24,7 @@ describe('github issues: Date field with cache style', () => {
         workbookReader.on('error', reject);
       })
   );
-  it('issue 1328 - should emit row with Date Object', () => {
+  it.skip('issue 1328 - should emit row with Date Object', () => {
```
gameroman, line 27: .

gameroman (changes requested):
Were these changes intended?

yuanliwei:
> Were these changes intended?

Thanks for the review. You're right, those changes (package name and the skipped test) were not intended for this PR — they were local modifications for my own fork. I'll close this and open a clean PR with only the fix and the test case. Sorry for the noise.

PR #3071, opened 2026-08-07
Fix workbook corruption when print titles or print areas reference sheet names containing apostrophes

pranshuchittora:
## Summary

Workbooks written by exceljs are reported as corrupt by Excel ("We found a problem with some content in '….xlsx'. Do you want us to try to recover as much as we can?") when a worksheet whose name contains an apostrophe has `pageSetup.printTitlesRow`, `printTitlesColumn`, or `printArea` set. Clicking "Yes" recovers the file (Excel drops the defined names), but automated pipelines consuming the files reject them outright.

Found in the same real-world workbook as #3070 — the two are independent corruption causes.

## Reproduction

```js
const wb = new ExcelJS.Workbook();
const ws = wb.addWorksheet("D'Alsace");
ws.getCell('A1').value = 'title row';
ws.pageSetup.printTitlesRow = '1:1';
await wb.xlsx.writeFile('repro.xlsx'); // Excel prompts for repair on open
```

The same applies to sheets read from an existing file: reading a template whose sheet has print titles, renaming the sheet to something containing an apostrophe, and writing it back produces the corrupt defined name.

## Root cause

`WorkbookXform.prepare()` rebuilds the `_xlnm.Print_Titles` / `_xlnm.Print_Area` defined names from `pageSetup`, interpolating the sheet name into the formula string:

```js
ranges.push(`'${sheet.name}'!$${titlesRows[0]}:$${titlesRows[1]}`);
```

In the formula grammar (ECMA-376), an apostrophe inside a quoted sheet name must be doubled. For a sheet named `D'Alsace` this emits `'D'Alsace'!$1:$1` instead of `'D''Alsace'!$1:$1`, which is not parseable as a sheet reference, so Excel flags the workbook part as corrupt. (#822 introduced the quoting itself but did not handle apostrophes within the name.)

## Fix

Escape apostrophes by doubling them at the three interpolation sites in `prepare()` (one for `Print_Area`, two for `Print_Titles`). Sheet names without apostrophes serialize byte-identically to before.

## Tests

- New integration spec `spec/integration/issues/defined-names-sheet-name-apostrophe.spec.js`: writes a workbook with print titles and a print area on a sheet named `D'Alsace` and asserts both defined names contain the doubled apostrophe. Fails against the previous code, passes with the fix.
- `npm run test:unit`: 883 passing (unchanged).

### DIFF
--- a/lib/xlsx/xform/book/workbook-xform.js
+++ b/lib/xlsx/xform/book/workbook-xform.js
@@ -47,12 +47,15 @@ class WorkbookXform extends BaseXform {
     const printAreas = [];
     let index = 0; // sheets is sparse array - calc index manually
     model.sheets.forEach(sheet => {
+      // Apostrophes inside a quoted sheet name must be doubled (ECMA-376 formula grammar);
+      // otherwise Excel reports the workbook as corrupt.
+      const escapedSheetName = sheet.name.replace(/'/g, '\'\'');
       if (sheet.pageSetup && sheet.pageSetup.printArea) {
         sheet.pageSetup.printArea.split('&&').forEach(printArea => {
           const printAreaComponents = printArea.split(':');
           const definedName = {
             name: '_xlnm.Print_Area',
-            ranges: [`'${sheet.name}'!$${printAreaComponents[0]}:$${printAreaComponents[1]}`],
+            ranges: [`'${escapedSheetName}'!$${printAreaComponents[0]}:$${printAreaComponents[1]}`],
             localSheetId: index,
           };
           printAreas.push(definedName);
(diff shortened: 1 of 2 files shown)


gameroman:
This package is not maintained anymore, I'd recommend trying out one of the more actively maintained forks

https://github.com/excel-js/exceljs

https://github.com/cjnoname/excelts

https://github.com/protobi/exceljs

You can also see a [comparison on npmx](https://npmx.dev/compare?packages=exceljs,@cj-tech-master/excelts,@excel.js/exceljs,@protobi/exceljs)

PR #3081, opened 2026-09-08
Fix incorrect function description

bleakley:
Summary: Fix the incorrect description of addRow.

Test Plan: Non-functional comment-only change.


lbesecker195 (approved):
`addRow` adds a single row; the plural wording was copied from `addRows`. Comment-only. LGTM.

